PR #2204, opened 2022-12-07
Issue/2114 crawler

tatjana-gartner2:
### Pull Request Checklist

#### References

#### Scope


packages/crawler/reports/2022-12-19_data-aggregated.json
```diff
@@ -0,0 +1,1400 @@
...
+        "https://login.porsche.com/login/de/de_DE",
+        "https://shop.porsche.com/de/de-DE",
+        "https://www.porsche.com/swiss/de"
+    ],
+    "aggregatedConsumedTagNames": {
+        "tagNames": {
+            "p-button": {
+                "amount": 20,
+                "hostPdsComponent": 0,
+                "slot": 0,
```
affilnost, line 14: it's not correct that p-button has no slots

affilnost: - rename "slot" to "children"

- change the logic of getting slot contents: get children instead of going into shadowDOM

packages/crawler/reports/login.porsche.com-login/2022-12-19_data-raw.json
```diff
@@ -0,0 +1,1088 @@
...
+        "2.17.0": {
+            "": [
+                {
+                    "p-content-wrapper": {
+                        "properties": {
+                            "width": "fluid",
+                            "backgroundColor": "transparent",
+                            "theme": "light"
+                        },
+                        "slot": "<myaccess-login-background _ngcontent-serverapp-c85=\"\" _nghost-serverapp-c84=\"\">...</myaccess-login-background><p-flex _ngcontent-serverapp-c85=\"\" direction=\"column\" class=\"site hydrated\">...</p-flex>"
```
affilnost, line 18: - add comment in the code (implementation), why we added this
- replace "..." with some special text, for example "(the content was stripped)"

packages/crawler/reports/www.porsche.com/2022-12-19_data-aggregated.json
```diff
@@ -0,0 +1,1130 @@
...
+                    "target": {
+                        "amount": 1,
+                        "values": {
+                            "_self": 1
+                        }
+                    },
+                    "aria": {
+                        "amount": 1,
+                        "values": {
+                            "[object Object]": 1
```
affilnost, line 228: it looks strange, needs to be checked

packages/crawler/reports/2022-12-19_data-aggregated.json
```diff
@@ -0,0 +1,1400 @@
...
+                            "false": 29,
+                            "true": 1
+                        }
+                    },
+                    "wrap": {
+                        "amount": 30,
+                        "values": {
+                            "nowrap": 25,
+                            "wrap": 4,
+                            "{base: 'wrap', xs: 'nowrap'}": 1
```
affilnost, line 193: it's not a valid json - we should use "JSON.stringify" in order to have a unique format

affilnost (commented):
TODO: use "getComponentMeta" and INTERNAL_TAG_NAMES

affilnost (commented):
**Done:** use "getComponentMeta" and INTERNAL_TAG_NAMES

affilnost (commented):
TODO: write minimum e2e test with sample html and check data_raw as snapshot

affilnost (commented):
- move index.ts to src folder
- split out websiteUrls from the general config and pass it as parameter in crawlWebsites
- PropValue Type - add object

packages/crawler/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "crawler",
+  "version": "1.0.0",
+  "description": "Porsche Design System Crawler",
+  "repository": "git@github.com:porsche-design-system/crawler.git",
+  "author": "Dr. Ing. h.c. F. Porsche AG",
+  "license": "MIT",
```
affilnost, line 7: - look in another package: license, private, version 0.0.0, repository
- remove prettier
- check devendencies, puppeteer is missing, shared package

packages/crawler/README.md
```diff
@@ -0,0 +1,44 @@
...
+- Aggregated data - amout of PDS components and properties
+- Information about PDS components' slot's contents
+- Information about parent PDS component
+
+## Available Scripts
+
+### Project setup
+
+```
+yarn install
```
affilnost, line 17: remove "yarn install"

affilnost: done

packages/crawler/reports/www.porsche.com/2022-12-19_data-aggregated.json
affilnost: fixed it

affilnost:
As discussed in Daily, I also corrected the ComponentMeta Type:

`[propName: string]: boolean | number | string | object | null;`

packages/crawler/reports/2022-12-19_data-aggregated.json
```diff
@@ -0,0 +1,1400 @@
...
+                    "theme": {
+                        "amount": 20,
+                        "values": {
+                            "light": 14,
+                            "dark": 6
+                        }
+                    }
+                },
+                "unusedProperties": [
+                    "isDelegatingFocus",
```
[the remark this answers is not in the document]
affilnost: - Fixed the bug
- If the default property value is null - it's value will be undefined after rendering, so we have unused values

packages/crawler/reports/login.porsche.com-login/2022-12-19_data-raw.json
affilnost: done

packages/crawler/reports/finder.porsche.com-de/2022-12-19_data-aggregated.json
```diff
@@ -0,0 +1,716 @@
...
+            "p-toast-item"
+        ]
+    },
+    "aggregatedConsumedTagNamesForVersionsAndPrefixes": {
+        "2.14.0": {
+            "fn": {
+                "tagNames": {
+                    "p-marque": {
+                        "amount": 1,
+                        "hostPdsComponent": 0,
```
affilnost: make null by default

packages/crawler/reports/2022-12-19_data-aggregated.json
affilnost: done

packages/crawler/reports/finder.porsche.com-de/2022-12-19_data-aggregated.json
affilnost: done

packages/crawler/reports/2022-12-19_data-aggregated.json
affilnost: Investigated is: it's the way it's implemented on https://www.porsche.com/swiss/de/. They put `{base: 'wrap', xs: 'nowrap'}` as a value of "size" attribute. The questions are:

- is it working like this (is the size still being applied, even if it's not a correct json)?
- if it's working - should I then convert this value to an object? (by making it to be correct json format)

TODO: discuss with the team

packages/crawler/package.json
affilnost: Done

packages/crawler/reports/2022-12-19_data-aggregated.json
affilnost: Discussed with the team: we should parse this object using existing functionality (parseJSON)

affilnost: Done

packages/crawler/tests/e2e/specs/index.e2e.ts
```diff
@@ -0,0 +1,85 @@
+import { crawlPage } from '../../../src/crawl-websites';
+import * as puppeteer from 'puppeteer';
+import { puppeteerConfig } from '../../../constants';
+import { PdsTestingContext, setContentWithDesignSystem } from '../helpers';
+
+export const testCrawlerWithHtmlAndPrefixes = async (pdsTestingContext: PdsTestingContext): Promise<void> => {
+  const browser = await puppeteer.launch(puppeteerConfig);
```
denyo, line 7: Can't we use `jest-puppeteer` with the setup of our shared package here?

packages/crawler/tests/e2e/helpers/puppeteer-helper.ts
```diff
@@ -0,0 +1,121 @@
...
+  const prefixesLoaders = prefixes.reduce(
+    (result, prefix) =>
+      result +
+      `
+    porscheDesignSystem.load(\{ prefix: '${prefix}' \});
+  `,
+    ''
+  );
+  // the script below has been copied from https://designsystem.porsche.com/v2/
+  return `
```
denyo, line 22: Perhaps we can get rid of the formatting. Might be related to prettier which could be disabled for a section or next line.

packages/crawler/tests/e2e/helpers/puppeteer-helper.ts
```diff
@@ -0,0 +1,121 @@
...
+export type PdsTestingContext = {
+  bodyHtml: string;
+  firstPdsVersionPrefixes?: string[];
+  secondPdsVersionPrefixes?: string[];
+};
+export const getInternalLoaderScriptForPrefixes = (prefixes: string[]): string =>
+  prefixes.reduce((result, prefix) => result + getLoaderScript({ prefix: prefix }), '');
+
+export const getExternalLoaderScriptForPrefixes = (prefixes: string[]): string => {
+  const prefixesLoaders = prefixes.reduce(
```
denyo, line 13: Could be replaced by `.map().join()` for less multiline string interpolation.

packages/crawler/tests/e2e/helpers/puppeteer-helper.ts
```diff
@@ -0,0 +1,121 @@
...
+          })();
+          ${prefixesLoaders}
+        </script>`;
+};
+
+export const setContentWithDesignSystem = async (page: Page, pdsTestingContext: PdsTestingContext): Promise<void> => {
+  const options: WaitForOptions = {
+    waitUntil: 'networkidle0',
+  };
+  const firstPdsVersionPrefixes = pdsTestingContext.firstPdsVersionPrefixes;
```
denyo, line 89: Could be destructured instead of aliased.

packages/crawler/tests/e2e/config/jest.config.ts
```diff
@@ -0,0 +1,31 @@
+const defaultWorkers = process.env.CI === 'true' ? 1 : 2;
```
denyo, line 1: Why duplicate instead of reusing the shared one?

packages/crawler/src/types.ts
```diff
@@ -0,0 +1,71 @@
+import { TagName } from 'shared/src';
```
denyo, line 1: Should be imported from `@porsche-design-system/shared`

packages/crawler/src/helpers/fs-helper.ts
```diff
@@ -0,0 +1,55 @@
...
+  );
+  fs.writeFileSync(
+    `./${config.reportFolderName}/${websiteName}/${new Date().toJSON().slice(0, 10)}${
+      config.dateSplitter
+    }data-aggregated.json`,
+    aggregatedData
+  );
+};
+
+export const writeGeneralReport = (aggregatedData: string) => {
```
denyo, line 29: Explicit return type is missing.

packages/crawler/src/helpers/fs-helper.ts
```diff
@@ -0,0 +1,55 @@
+import { crawlerConfig as config } from '../../constants';
+import fs from 'fs';
+
+const getWebsiteNameByWebsiteUrl = (websiteUrl: string): string => {
```
denyo, line 4: Return type is missing, too.

packages/crawler/src/helpers/convert-data-helper.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export const getPdsTagNamesWithPropertyNames = (): TagNamesWithPropertyNames =>
+  Object.values(TAG_NAMES).reduce((result, tagName) => {
+    const componentMeta = getComponentMeta(tagName);
+    return !INTERNAL_TAG_NAMES.includes(tagName)
+      ? {
+          ...result,
+          [tagName]: componentMeta.props ? Object.keys(componentMeta.props) : [],
+        }
+      : { ...result };
```
denyo, line 20: Unnecessary new object.

packages/crawler/src/helpers/convert-data-helper.ts
```diff
@@ -0,0 +1,79 @@
+import {
```
denyo, line 1:
```suggestion
import type {
```

packages/crawler/src/evaluate-page.ts
```diff
@@ -0,0 +1,155 @@
...
+      };
+
+      const getAllChildElements = (el: Element): Element[] => {
+        const children = Array.from(el.children)
+          .concat(Array.from(el.shadowRoot?.children || []))
+          .flat() as Element[];
+        return children.concat(children.map(getAllChildElements).flat());
+      };
+
+      const getChildren = (el: Element): string | null =>
```
denyo, line 37: Function name doesn't really match what's happening?

packages/crawler/src/evaluate-page.ts
```diff
@@ -0,0 +1,155 @@
...
+        Array.from(el.children).reduce((result, slotEl) => {
```
denyo, line 38: Shouldn't `slotEl` be reamed to `child`?

packages/crawler/src/evaluate-page.ts
```diff
@@ -0,0 +1,155 @@
...
+            return result + copy.outerHTML;
+          }
+          return result + slotEl.outerHTML;
+        }, '' as string) || el.textContent;
+
+      const getHostPdsComponent = (el: Element, prefix: string): TagName | null => {
+        const rootNode = (el.getRootNode() as ShadowRoot).host;
+        if (rootNode) {
+          return getPdsComponentNameByPrefix(rootNode, prefix);
+        }
```
denyo, line 52: Move `return null` into  else case for clarity.

packages/crawler/src/evaluate-page.ts
```diff
@@ -0,0 +1,155 @@
...
+          try {
+            // jsonStr is potentially JSON parsable string, e.g. "{ base: 'block', l: 'inline' }"
+            return JSON.parse(
+              jsonStr
+                .replace(/'/g, '"') // convert single quotes to double quotes
+                .replace(/[\s"]?([a-z\-]+)[\s"]?:([^//])/g, '"$1":$2') // wrap keys in double quotes if they don't have them but ignore potential urls
+            );
+          } catch {
+            // jsonStr is string, e.g. "block" or "inline"
+            return jsonStr;
```
denyo, line 67: Probably a dead brach because of 2nd condition: `&& jsonStr.includes('{')`.

packages/crawler/src/evaluate-page.ts
```diff
@@ -0,0 +1,155 @@
...
+
+      const getAllConsumedProperties = <
+        PComponentName extends keyof HTMLElementTagNameMap,
+        PComponentElement extends HTMLElementTagNameMap[PComponentName],
+        PComponentPropertyName extends keyof PComponentElement,
+        PComponentPropertyValue extends keyof PComponentElement[PComponentPropertyName]
+      >(
+        el: Element,
+        allPdsPropertiesForComponentName: string[]
+      ) => {
```
denyo, line 83: Return type is missing.

packages/crawler/src/evaluate-page.ts
```diff
@@ -0,0 +1,155 @@
...
+        const pEl = el as PComponentElement;
```
denyo, line 84: Alias might lead to confusion.

packages/crawler/src/evaluate-page.ts
```diff
@@ -0,0 +1,155 @@
...
+          if (!componentName) {
+            throw new Error('Could not find component name');
+          }
+
+          const slotInfo = getChildren(el);
+          const hostPdsComponent = getHostPdsComponent(el, prefix);
+
+          return {
+            [componentName]: {
+              ...{
```
denyo, line 116: Unnecessary object composition.

packages/crawler/src/evaluate-page.ts
```diff
@@ -0,0 +1,155 @@
...
+                properties: getAllConsumedProperties(el, pdsTagNamesWithPropertyNames[componentName]),
+              },
+              ...(slotInfo ? { children: slotInfo } : {}),
```
denyo, line 119: Could be simplified:
```suggestion
              children: slotInfo || {},
```

denyo, line 119: Could be simplified:
```suggestion
              children: slotInfo || {},
```

packages/crawler/src/crawl-websites.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  // aggregated data
+  const aggregatedConsumedTagNamesForVersionsAndPrefixes =
+    getAggregatedConsumedTagNamesForVersionsAndPrefixes(pdsCrawlerRawData);
+  // aggregated data without versions and prefixes
+  const aggregatedConsumedTagNames = getAggregatedConsumedTagNames(pdsCrawlerRawDataWithoutVersionsAndPrefixes);
+
+  writeWebsiteReport(
+    websiteUrl,
+    JSON.stringify(
```
denyo, line 46: Perhaps a util for this would make sense since the config is duplicated multiple times.

packages/crawler/src/crawl-websites.ts
```diff
@@ -0,0 +1,95 @@
...
+  getAggregatedConsumedTagNamesForVersionsAndPrefixes,
+  getConsumedPrefixesForVersions,
+  getPdsTagNamesWithPropertyNames,
+  getRawDataWithoutVersionsAndPrefixes,
+} from './helpers/convert-data-helper';
+import { writeGeneralReport, writeWebsiteReport } from './helpers/fs-helper';
+import { ConsumedTagNamesForVersionsAndPrefixes, TagNameData, TagNamesWithPropertyNames } from './types';
+import { Page } from 'puppeteer';
+
+export const crawlPage = async (page: Page, websiteUrl: string): Promise<ConsumedTagNamesForVersionsAndPrefixes> => {
```
denyo, line 15: I guess this was extracted to be used in the e2e test which makes sense.
However, naming is a bit misleading with `crawlPage`, `crawlWebsite` and `crawlWebsites`.

package.json
```diff
@@ -56,9 +56,11 @@
     "start:components-angular": "yarn --cwd ./packages/components-angular start",
     "start:components-react": "yarn --cwd ./packages/components-react start",
     "start:components-vue": "yarn --cwd ./packages/components-vue start",
+    "start:crawler": "yarn --cwd ./packages/crawler start",
     "start:fallbacks": "yarn --cwd ./packages/fallbacks start",
     "start:storefront": "yarn --cwd ./packages/storefront start",
     "start:utilities-deprecated": "yarn --cwd ./packages/utilities-deprecated start",
+    "start:assets": "yarn --cwd ./packages/assets start",
```
denyo, line 63: Let's revert this since it doesn't make a lot of sense to start the cdn like this.

package.json
```diff
@@ -108,6 +110,7 @@
     "test:e2e:components-react": "yarn --cwd ./packages/components-react test:e2e",
     "test:e2e:components-vue": "yarn --cwd ./packages/components-vue test:e2e",
     "test:e2e:storefront": "yarn --cwd ./packages/storefront test:e2e",
+    "test:e2e:crawler": "yarn --cwd ./packages/crawler test:e2e",
```
denyo, line 113: Please stick to the same order. The `start` command was listed after `components-vue`, but here it is after `storefront`.

packages/crawler/README.md
```diff
@@ -0,0 +1,43 @@
...
+- Used PDS components and configuration per component
+- PDS components and properties which are never used
+- Aggregated data - amout of PDS components and properties
+- Information about PDS components' slot's contents (children)
+- Information about parent PDS component
+
+## Start Crawler
+
+```
+yarn start:crawler
```
denyo, line 15: Isn't it just `yarn start` from this location?

packages/crawler/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "crawler",
+  "version": "0.0.0",
+  "private": true,
+  "license": "UNLICENSED",
+  "scope": "@porsche-design-system",
+  "author": "Dr. Ing. h.c. F. Porsche AG",
+  "scripts": {
+    "start": "ts-node src/index.ts",
+    "test:e2e:puppeteer": "jest --colors -c tests/e2e/config/jest.config.ts",
```
denyo, line 10: This could be simplified by using the puppeteer setup from our shared package. There, by passing a `PORT` env variable the setup is going to execute `yarn start-app` and wait for that port to be ready.
See other packages for reference.

packages/crawler/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "test:e2e": "concurrently -k -s 'first' -n 'cdn,puppeteer' 'serve-cdn' 'yarn test:e2e:puppeteer'"
+  },
+  "dependencies": {
```
denyo, line 13: For our licensing (OSO) scan the convention is that `dependencies` are dependencies that are bundled/shipped. While this package might be shipped/used in the future we should have this clean already, otherwise everything should be under `devDependencies`.  
- Therefore `typescript` and `ts-node` will never be shipped since those are used at build time.
- `puppeteer` most likely won't be shipped either and is in fact a part of `@porsche-design-system/shared` already and can therefore be removed.
- At the same time `jest` is missing but is provided by `@porsche-design-system/shared`
- `@porsche-design-system/components-js` should have a proper version which should also be updated in the `run-prepare-release-components` bash script.

packages/crawler/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "name": "crawler",
+  "version": "0.0.0",
+  "private": true,
+  "license": "UNLICENSED",
+  "scope": "@porsche-design-system",
+  "author": "Dr. Ing. h.c. F. Porsche AG",
+  "scripts": {
+    "start": "ts-node src/index.ts",
+    "test:e2e:puppeteer": "jest --colors -c tests/e2e/config/jest.config.ts",
+    "test:e2e": "concurrently -k -s 'first' -n 'cdn,puppeteer' 'serve-cdn' 'yarn test:e2e:puppeteer'"
```
denyo, line 11: `clean` commands are missing and should be called from root `package.json`, too.

denyo:
In addition to the comments the CI integration is currently missing or was that left out on purpose?

packages/crawler/package.json
```diff
@@ -0,0 +1,21 @@
...
+  },
+  "dependencies": {
+    "@porsche-design-system/shared": "0.0.0",
+    "@porsche-design-system/components-js": "0.0.0",
+    "puppeteer": "~19.3.0",
+    "ts-node": "^10.9.1",
+    "typescript": "~4.9.3"
+  },
+  "devDependencies": {}
```
denyo, line 20: `@porsche-design-system/assets` package is missing and necessary to provide the `serve-cdn` executable.

packages/crawler/tests/e2e/helpers/puppeteer-helper.ts
affilnost: Fixed

package.json
affilnost: true, fixed

packages/crawler/README.md
affilnost: true, fixed

packages/crawler/package.json
affilnost: Fixed

packages/crawler/src/types.ts
affilnost: Fixed

packages/crawler/src/helpers/fs-helper.ts
affilnost: I guess you meant "writeWebsiteReport" function - fixed it there too

packages/crawler/src/helpers/convert-data-helper.ts
affilnost: Fixed

packages/crawler/package.json
affilnost: @denyo I saw such implementation in other packages, but in this case we don't need "start-app", because we don't need any app (web server) to be running for e2e test, only cdn.. Therefore, we don't need any port to be passed and we can't wait till port is ready

packages/crawler/package.json
affilnost: Added this package

packages/crawler/package.json
affilnost: Done

packages/crawler/src/evaluate-page.ts
affilnost: @denyo In a case there's no "slotInfo" or no "hostPdsComponent", I want the returned object not to have the keys "children" and "hostPdsComponent", therefore I use object composition here..

Or is it possible to do it without object composition?

packages/crawler/src/crawl-websites.ts
affilnost: removed "crawlPage" and added calling "evaluatePage" directly from e2e tests

packages/crawler/src/crawl-websites.ts
affilnost: Added util for this

packages/crawler/src/evaluate-page.ts
affilnost: true, renamed the function

packages/crawler/src/evaluate-page.ts
affilnost: Renamed it to child

packages/crawler/src/evaluate-page.ts
affilnost: Done

packages/crawler/src/evaluate-page.ts
affilnost: It was covering the cases like "true" or "123", but actually it's better if we parse them too, so removed `&& jsonStr.includes('{')`

packages/crawler/src/evaluate-page.ts
affilnost: Fixed

packages/crawler/src/evaluate-page.ts
affilnost: true.. I was also thinking that it might lead to confusion.. But this alias is actually logical, because we take an Element and consider it to be "PDS Component" with corresponding properties.. Added comment in code to make it more clear:
`// we consider this element as PDS Component, therefore we need this alias here`

packages/crawler/src/helpers/convert-data-helper.ts
affilnost: true, fixed

packages/crawler/src/helpers/fs-helper.ts
```diff
@@ -0,0 +1,55 @@
...
+    aggregatedData
+  );
+};
+
+const removeOutdatedReportsInFolder = (folderName: string): void => {
+  fs.readdirSync(folderName)
+    .filter(
+      (fileName: string) => Date.parse(fileName.split(config.dateSplitter)[0]) < Date.now() - config.reportsMaxAge
+    )
+    .map((fileName: string) => {
```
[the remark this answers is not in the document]
affilnost: Fixed

packages/crawler/tests/e2e/helpers/puppeteer-helper.ts
affilnost: Fixed

packages/crawler/package.json
affilnost: @denyo I thought about it the second time, and I will try using shared package config, just without PORT (since we don't have web server running)

packages/crawler/src/helpers/fs-helper.ts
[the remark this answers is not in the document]
affilnost: Fixed

packages/crawler/tests/e2e/specs/index.e2e.ts
```diff
@@ -0,0 +1,85 @@
...
+  await setContentWithDesignSystem(page, pdsTestingContext);
+  const crawlerData = await crawlPage(page, 'http://localhost');
+  await browser.close();
+
+  // check that raw data matches snapshot
+  expect(crawlerData).toMatchSnapshot();
+};
+
+it('should retrieve children and hostPdsComponent correctly', async () => {
+  await testCrawlerWithHtmlAndPrefixes({
```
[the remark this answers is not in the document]
affilnost: Done

packages/crawler/package.json
affilnost: I changed the implementation - now we use shared package config. We don't pass PORT here, because we don't need the web server.

packages/crawler/src/evaluate-page.ts
affilnost: I changed "children" and "hostPdsComponent" to be null by default, so could get rid of composition here.

packages/crawler/src/evaluate-page.ts
affilnost: Done

packages/crawler/tests/e2e/config/jest.config.ts
affilnost: Fixed it - now we use shared config

packages/crawler/tests/e2e/specs/index.e2e.ts
affilnost: Fixed it

package.json
```diff
@@ -122,7 +124,7 @@
...
+    "clean:node_modules": "rm -rf ./node_modules && rm -rf ./yarn-error.log && yarn clean:assets:node_modules && yarn clean:components:node_modules && yarn clean:components-angular:node_modules && yarn clean:components-js:node_modules && yarn clean:components-react:node_modules && yarn clean:components-vue:node_modules && yarn clean:crawler:node_modules && yarn clean:email:node_modules && yarn clean:fonts:node_modules && yarn clean:icons:node_modules && yarn clean:marque:node_modules && yarn clean:meta-icons:node_modules && yarn clean:shared:node_modules && yarn clean:storefront:node_modules && yarn clean:utilities:node_modules && yarn clean:utilities-deprecated:node_modules",
     "clean:build": "yarn clean:assets:build && yarn clean:components:build && yarn clean:components-angular:build && yarn clean:components-js:build && yarn clean:components-react:build && yar clean:components-vue:build && yarn clean:email:build && yarn clean:fonts:build && yarn clean:icons:build && yarn clean:marque:build && yarn clean:meta-icons:build && yarn clean:shared:build && yarn clean:storefront:build && yarn clean:utilities:build && yarn clean:utilities-deprecated:build",
```
stephanschroeter, line 128: Is there nothing to clean regarding builds?

docker/node/bin/run-prepare-release-components
```diff
@@ -51,6 +51,11 @@ for pkg in ${packages[@]}; do
   popd
 done
 
+# update "components-js" in crawler project
+pushd "./packages/crawler"
```
denyo, line 55: Should be added to array of L60.

docker/node/bin/run-test-e2e
```diff
@@ -7,7 +7,8 @@ if
   [[ "--components-angular" == "${1}" ]] ||
   [[ "--components-react" == "${1}" ]] ||
   [[ "--components-vue" == "${1}" ]] ||
-  [[ "--storefront" == "${1}" ]]
+  [[ "--storefront" == "${1}" ]] ||
```
denyo, line 10: If we would stick to the same order everywhere, then crawler should come after components-vue, right?

packages/crawler/constants.ts
```diff
@@ -0,0 +1,11 @@
+import { CrawlerConfig } from './src/types';
```
denyo, line 1:
```suggestion
import type { CrawlerConfig } from './src/types';
```

packages/crawler/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "scope": "@porsche-design-system",
+  "author": "Dr. Ing. h.c. F. Porsche AG",
+  "scripts": {
+    "start": "ts-node src/index.ts",
+    "test:e2e:puppeteer": "jest --colors -c tests/e2e/config/jest.config.ts",
+    "test:e2e": "concurrently -k -s 'first' -n 'cdn,puppeteer' 'serve-cdn' 'yarn test:e2e:puppeteer'",
+    "clean:node_modules": "rm -rf ./node_modules && rm -rf ./yarn-error.log"
+  },
+  "dependencies": {
+    "@porsche-design-system/shared": "0.0.0",
```
denyo, line 15: Please order alphabetically.

packages/crawler/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "version": "0.0.0",
+  "private": true,
+  "license": "UNLICENSED",
+  "scope": "@porsche-design-system",
+  "author": "Dr. Ing. h.c. F. Porsche AG",
+  "scripts": {
+    "start": "ts-node src/index.ts",
+    "test:e2e:puppeteer": "jest --colors -c tests/e2e/config/jest.config.ts",
+    "test:e2e": "concurrently -k -s 'first' -n 'cdn,puppeteer' 'serve-cdn' 'yarn test:e2e:puppeteer'",
+    "clean:node_modules": "rm -rf ./node_modules && rm -rf ./yarn-error.log"
```
denyo, line 12: `clean:node_modules` is a child command, so there should be a parent command called `clean`.

packages/crawler/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "crawler",
+  "version": "0.0.0",
+  "private": true,
+  "license": "UNLICENSED",
+  "scope": "@porsche-design-system",
+  "author": "Dr. Ing. h.c. F. Porsche AG",
+  "scripts": {
+    "start": "ts-node src/index.ts",
+    "test:e2e:puppeteer": "jest --colors -c tests/e2e/config/jest.config.ts",
```
denyo, line 10: I think we can drop that child command and just do its stuff inline in the current parent command.

packages/crawler/package.json
denyo: I suppose with concurrently starting the cdn and executing the tests in parallel everything works. 
However this will probably kill interactive testing but definitely destroy passing additional parameters.  
A way around this could be to pass the `PORT` env variable and add a `start-app` command which then calls `serve-cdn`?

packages/crawler/constants.ts
```diff
@@ -0,0 +1,11 @@
...
+export const crawlerConfig: CrawlerConfig = {
+  reportFolderName: 'reports',
+  dateSplitter: '_',
+  jsonSpace: 4,
+  reportsMaxAge: 1000 * 60 * 60 * 24 * 7, // one week
```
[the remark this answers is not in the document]
affilnost: @stephanschroeter Reports are not checked in, but if you run the crawler locally and you have some old outdated reports, then they will be removed locally.. But maybe we don't need this functionality, I don't know...

package.json
affilnost: @stephanschroeter We don't have any builds right now, so no builds to be removed.. Removing reports during clean is a good idea.. But if we do it - then we indeed don't need "removing outdated reports" functionality in crawler

packages/crawler/constants.ts
affilnost: Fixed

packages/crawler/package.json
affilnost: Fixed

packages/crawler/package.json
affilnost: Fixed

docker/node/bin/run-prepare-release-components
affilnost: Fixed

packages/crawler/package.json
affilnost: Changed it

packages/crawler/tests/e2e/specs/index.e2e.ts
```diff
@@ -0,0 +1,100 @@
...
+    bodyHtml: `
+        <p-banner theme="dark">
+          <span slot="title">Some notification title</span>
+          <span slot="description">
+            Some notification description. And some <a href="https://www.porsche.com/">LINK</a> element.
+          </span>
+        </p-banner>
+    `,
+  });
+  console.log('Evaluating page..');
```
denyo, line 20: Let's reduce these logs since they don't bring any value.
Bildschirm­foto 2023-01-17 um 09 09 33

.github/workflows/push.yml
```diff
@@ -579,6 +579,51 @@ jobs:
       with:
         name: components-vue-vrt-regression
         path: ./packages/components-vue/tests/vrt/results
+  crawler-e2e-puppeteer:
```
denyo, line 582: This job should probably become a condition for other jobs to start? You can search for `needs:` within this file to find some.

affilnost: Added it as a condition for Storefront and Assets deployments

packages/crawler/tests/e2e/specs/index.e2e.ts
affilnost: removed these logs

packages/crawler/constants.ts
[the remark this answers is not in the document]
affilnost: Removed Max Age and "removing outdated reports" functionality

package.json
affilnost: Added "remove reports" to clean

packages/components-react/projects/jsdom-polyfill/src/lib/a11y-4587e563.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
```
stephanschroeter, line 1: These files within jsdom-polyfill shouldn't be committed.

affilnost: Removed these files, they were committed occasionally during merge

package.json
```diff
@@ -125,7 +127,7 @@
...
+    "clean:node_modules": "rm -rf ./node_modules && rm -rf ./yarn-error.log && yarn clean:assets:node_modules && yarn clean:components:node_modules && yarn clean:components-angular:node_modules && yarn clean:components-js:node_modules && yarn clean:components-react:node_modules && yarn clean:components-vue:node_modules && yarn clean:crawler && yarn clean:email:node_modules && yarn clean:fonts:node_modules && yarn clean:icons:node_modules && yarn clean:marque:node_modules && yarn clean:meta-icons:node_modules && yarn clean:shared:node_modules && yarn clean:storefront:node_modules && yarn clean:utilities:node_modules && yarn clean:utilities-deprecated:node_modules",
```
denyo, line 130: child command should be called `clean:crawler:node_modules` to match the rest.

package.json
```diff
@@ -139,6 +141,7 @@
     "clean:components-react:build": "yarn --cwd ./packages/components-react clean:build",
     "clean:components-vue:node_modules": "yarn --cwd ./packages/components-vue clean:node_modules",
     "clean:components-vue:build": "yarn --cwd ./packages/components-vue clean:build",
+    "clean:crawler": "yarn --cwd ./packages/crawler clean",
```
denyo, line 144:
```suggestion
    "clean:crawler:node_modules": "yarn --cwd ./packages/crawler clean:node_modules",
```

stephanschroeter (approved):

packages/crawler/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "license": "UNLICENSED",
+  "scope": "@porsche-design-system",
+  "author": "Dr. Ing. h.c. F. Porsche AG",
+  "scripts": {
+    "start": "ts-node src/index.ts",
+    "start-app": "serve-cdn",
+    "test:e2e": "PORT=3001 jest --colors -c tests/e2e/config/jest.config.ts",
+    "clean": "yarn clean:node_modules && yarn clean:reports",
+    "clean:node_modules": "rm -rf ./node_modules && rm -rf ./yarn-error.log",
+    "clean:reports": "rm -rf ./reports"
```
denyo, line 14: Might be a good idea to call this upon `start`?

affilnost: good idea, changed it

package.json
affilnost: Changed it

package.json
affilnost: Changed it

Issue #2062, opened 2022-09-22
Modal component is not visually centered covering the viewport in preview mode


marcelbertram:
Scope
---
The PDS modal component needs to be placed on the canvas covering the whole application. When overlaying longer pages with scroll sections this causes that the modal is not visually centered in the viewport in preview mode.

Notes  
---
- Similar problem to the #1957 absolute positioning problem
- UX Pin Example: https://app.uxpin.com/edit/9058405#?id_page=158774852
- Problem with with different handling of canvas and browser window (editor and preview)
- Priority: Correct behavior in the preview simulate mode
- Use React Portal, see: https://recordit.co/uHa0YFGaal

Acceptance criteria
---
- [ ] Modal overlay spans the complete viewport
- [ ] Solution is also applied to toast component absolute positioning

Subtasks
---
- [ ] Task

PR #2208, opened 2022-12-14
Improve rendering of `Modal` and `Toast` using Merge "Portal" feature

michaelrambeau:
### Pull Request Checklist

- [X] I have read and accept the [Contributor License Agreement](https://github.com/porscheofficial/oss-docs/blob/67c5b24838a293ce7a964884e6005eb71f2b8579/CONTRIBUTOR_LICENSE_AGREEMENT.md)

### References

Fixes https://github.com/porsche-design-system/porsche-design-system/issues/2062 

The goal is to improve the rendering of components whose position on the screen is absolute (a `Modal` or a `Toast` component) by using a new feature from Merge component: the "Portal feature".

Portal components should be rendered at a position related to the screen instead of the position where they have been placed in the UXPin editor.

### Scope

This PR upgrades the Merge CLI dependency to the latest version (`2.9.0`) to detect a new keyword `@uxpinuseportal` when parsing components code. In UXPIn Preview, this keyword is used to render the component in a [React.Portal](https://reactjs.org/docs/portals.html#usage) instead of the usual canvas.

`@uxpinuseportal` is used in the optional _JSDoc_ comments located before the component declaration.

A basic example:

```tsx
/**
 * @uxpinuseportal
 */
export const Toast = ({ position, message }: Props) => {
  return (
    <div>
      {message}
    </div>
  );
};
```

This new keyword is not documented yet, but you can find the list of other available keywords here:

https://www.uxpin.com/docs/merge/adjusting-components/#mark-components-by-comment

For now, the two following components use the portal feature:

- `Modal`
- `Toast`
 
For the records, [this PR](https://github.com/UXPin/uxpin-merge-tools/pull/346) introduced the new feature in the CLI code base.

### How to check

Check the component behavior by launching the "Merge experimental mode" from `/packages/components-react/projects/uxpin-wrapper` folder: `yarn start`

#### `<Modal>`

- Add a Modal component to the canvas
- Set the property `Open` to false to hide the modal
- Add a button with an interaction that sets the property `Open` to `True`
- Launch the Preview mode
- Click on the button
- The modal should covert the main area of the Preview page (not the full screen to avoid covering UXPin controls)
 
image

#### `<Toast>`

- Add a Toast component, leaving is text property empty, anywhere on the canvas, it does not matter!
- Add a button with an interaction that sets the `Text` property to something like `Hello Toast`
- Launch the Preview mode
- Click on the button
- The toast component should be at the bottom of the screen.
 
image

#### Video

https://share.getcloudapp.com/Koune1OD


packages/components/scripts/wrapper-generator/UXPinReactWrapperGenerator.ts
```diff
@@ -440,6 +442,39 @@ export class UXPinReactWrapperGenerator extends ReactWrapperGenerator {
...
+  }
+
+  private getAllComponentComments(component: TagName): string[] {
+    const comments: string[] = [];
+
+    if (this.shouldRenderInReactPortal(component)) {
+      comments.push('* @uxpinuseportal');
+    }
+    return comments;
+  }
```
michaelrambeau, line 466: Now only the `@uxpinportal` feature is supported but my assumption was that we could support more properties from this list:

```
@uxpincomponent
@uxpinnamespace
@uxpinwrappers
@uxpindocurl
@uxpinuseportal
```
 
Of course the implementation could be even simpler if we consider that we only care about `@uxpinuseportal` 😅

packages/components-react/projects/uxpin-wrapper/package.json
```diff
@@ -23,7 +23,7 @@
   "devDependencies": {
     "@babel/preset-typescript": "^7.18.6",
     "@types/react": "^18.0.25",
-    "@uxpin/merge-cli": "2.8.0-dev.77",
+    "@uxpin/merge-cli": "2.9.0",
```
denyo, line 26: Please also update `docs/dependencies.md` by removing the entry related to this.

packages/components/scripts/wrapper-generator/UXPinReactWrapperGenerator.ts
```diff
@@ -440,6 +442,39 @@ export class UXPinReactWrapperGenerator extends ReactWrapperGenerator {
     return [...componentPresetFiles, configFile];
   }
 
+  // Component declaration can be preceded by JSDoc comments
+  // to customize the behavior in UXPin Editor or Preview (E.g.: render in a React Portal)
+  // https://www.uxpin.com/docs/merge/adjusting-components/
+  private insertComponentAnnotation(cleanedComponent: string, component: TagName): string {
+    const comments = this.getAllComponentComments(component);
+    if (!comments.length) return cleanedComponent;
```
denyo, line 450: Let's please avoid same line return and add an `else` case for better readability.

packages/components/scripts/wrapper-generator/UXPinReactWrapperGenerator.ts
```diff
@@ -440,6 +442,39 @@ export class UXPinReactWrapperGenerator extends ReactWrapperGenerator {
...
+*/
+`;
+    return annotations + cleanedComponent;
+  }
+
+  private getAllComponentComments(component: TagName): string[] {
+    const comments: string[] = [];
+
+    if (this.shouldRenderInReactPortal(component)) {
+      comments.push('* @uxpinuseportal');
```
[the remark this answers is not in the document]
michaelrambeau: That was my first move actually 😅 until I realized that there are other JSDoc comments that apply at the component level in UXPin.
OK, we may never need them so let's keep it as simple as possible 👍

packages/components/scripts/wrapper-generator/UXPinReactWrapperGenerator.ts
michaelrambeau: Done here https://github.com/porsche-design-system/porsche-design-system/pull/2208/commits/959c331a17d0b714cfaf3aec8ccd5b04442b42b0

I'm always happy to clean stuff 😄

packages/components-react/projects/uxpin-wrapper/package.json
michaelrambeau: Done here https://github.com/porsche-design-system/porsche-design-system/pull/2208/commits/959c331a17d0b714cfaf3aec8ccd5b04442b42b0

I'm always happy to clean stuff 😄

denyo (approved):

PR #2251, opened 2023-01-19
Issue/2149 tag v3

affilnost:
### Pull Request Checklist

#### References

#### Scope


packages/components-js/tests/unit/fixtures/stats.json
```diff
@@ -1,207 +1,207 @@
...
-    "size": 3912,
-    "gzipSize": 1720
+    "size": 4684,
+    "gzipSize": 1857
   },
   {
     "chunkShortName": "tag-dismissible",
-    "size": 5193,
-    "gzipSize": 2134
+    "size": 4610,
```
affilnost, line 179: FYI: tag-dismissible chunk size decreased because I moved some code from `tag-styles.ts` into `tag-shared-utils.ts`, so that changing `tag-styles.ts` doesn't affect the "tag-dismissible" chunk size

packages/components/src/components/tag/tag-shared-utils.ts
```diff
@@ -0,0 +1,45 @@
...
+      borderColor: 'transparent',
+    },
+  };
+};
+
+export const getThemedBackgroundColor = (tagColor: TagColor, themedColors: ThemedColors): string => {
+  const colorMap: { [key in TagColor]: string } = {
+    'background-default': themedColors.backgroundColor,
+    'background-surface': themedColors.backgroundSurfaceColor,
+    'neutral-contrast-high': themedColors.primaryColor,
```
affilnost, line 28: FYI: As discussed with Janick and Johannes, they want the background color for 'neutral-contrast-high' to be primaryColor (black), because the 'contrastHighColor' is not dark enough

packages/components/src/components/tag/tag-shared-utils.ts
```diff
@@ -0,0 +1,45 @@
+import { getInsetJssStyle, ThemedColors } from '../../styles';
```
affilnost, line 1: I created this file in order to place here all the constants which are being used in "tag-dismissible", so that changing "tag-styles.ts" doesn't affect "tag-dismissible"'s chunk size

packages/components/scripts/generateThemesObject.ts
```diff
@@ -7,6 +7,10 @@ import tinycolor2 from 'tinycolor2';
 import { pascalCase } from 'change-case';
 
 const darkenColor = (color: string) => tinycolor2(color).darken(12).toHexString();
```
affilnost, line 9: FYI:

'darkenColor' is being used for 'hover background' for light theme, no colorful (i.e. backgroundColor or contrastHighColor)
'darkenColorSlightly' is being used for 'hover backround' for light theme, colorful (i.e. success, error, warning, information)

'lightenColor' - the same as 'darkenColor', only for dark theme
'lightenColorSlightly' - the same as 'darkenColorSlightly', only for dark theme

packages/components/src/components/tag/tag-shared-utils.ts
stephanschroeter: Hmm, I see the point but I don't like the inconsistency. Why not adding in addition a new accepted color value `primary` and deprecate `neutral-contrast-high` while it's being mapped to `primary` for backwards compatibility reasons?

Same applies to `background-default`, which should be aliased to `background-base`.

packages/components/src/components/tag/tag-shared-utils.ts
```diff
@@ -0,0 +1,45 @@
...
+    'notification-information': themedColors.infoSoftColor,
+    'notification-success': themedColors.successSoftColor,
+    'notification-error': themedColors.errorSoftColor,
+    'notification-warning': themedColors.warningSoftColor,
+  };
+
+  return colorMap[tagColor];
+};
+
+export const slottedTextJssStyle: JssStyle = {
```
stephanschroeter, line 38: I think we should remove those custom slotted styles completely because we'll handle those generic light dom styles with the new normalize styles provided by the partial `getInitialStyles()` in the future.

packages/components/src/components/tag/tag-styles.spec.ts
```diff
@@ -60,27 +37,29 @@ xdescribe('getColors()', () => {
...
+describe('getComponentCss()', () => {
   it.each<Parameters<typeof getComponentCss>>([
     ['background-default', true, 'light'],
     ['background-default', false, 'light'],
     ['neutral-contrast-high', true, 'light'],
     ['notification-success', true, 'light'],
     ['notification-warning', true, 'light'],
     ['notification-error', true, 'light'],
-    ['notification-neutral', true, 'light'],
+    ['notification-information', true, 'light'],
```
stephanschroeter, line 54: `notification-neutral` should be deprecated but still being supported and mapped to `notification-information` for backwards compatibility reasons.

packages/components/src/components/tag/tag-styles.ts
```diff
@@ -100,21 +52,20 @@ export const getComponentCss = (tagColor: TagColor, isFocusable: boolean, theme:
...
+        padding: '5px 9px',
+        borderRadius: borderRadiusSmall,
         background: backgroundColor,
         color: primaryColor,
         ...textXSmallStyle,
         whiteSpace: 'nowrap',
-        ...(isFocusable && {
-          transition: getTransition('color'),
-          ...hoverMediaQuery({
+        transition: `${getTransition('background-color')}`,
```
stephanschroeter, line 61: backticks are unnecessary here

stephanschroeter: In addition, shouldn't it be part of the `isFocusable` condition?

packages/components/src/components/tag/tag-utils.ts
```diff
@@ -1,11 +1,12 @@
 import { TAG_DISMISSIBLE_COLORS } from '../tag-dismissible/tag-dismissible-utils';
 import type { Theme } from '../../types';
-import { isThemeDark } from '../../utils/theme/isThemeDark'; // deep import needed since barrel contains MutationObserver and causes VRT to fail because of TAG_COLORS import
+import { isThemeDark } from '../../utils/theme/isThemeDark';
+import { ThemedColors } from '../../styles'; // deep import needed since barrel contains MutationObserver and causes VRT to fail because of TAG_COLORS import
```
stephanschroeter, line 4: since it's a type import, it should be declared as such.

stephanschroeter (changes requested):
Newly introduced scaling test is missing in `packages/components-js/tests/vrt/puppeteer/specs/component-scaling.vrt.ts`

packages/components/CHANGELOG.md
```diff
@@ -16,7 +16,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 #### Changed
 
-- `Spinner`, `Link Pure`, `Button Pure`, `Checkbox Wrapper`, `Radio Button Wrapper`, `Tag Dismissible` matches new design language
+- `Spinner`, `Link Pure`, `Button Pure`, `Checkbox Wrapper`, `Radio Button Wrapper`, `Tag Dismissible`, `Tag` matches
+  new design language
+- Color `notification-neutral` is renamed to `notification-information` for `Tag`
```
denyo, line 21: Please add a diff for breaking changes so people know what to do.

packages/components/src/components/tag/tag-utils.ts
denyo: I think comment is on the wrong line now?

packages/components/src/components/tag/tag-utils.ts
```diff
@@ -16,6 +17,29 @@ export const hasInvertedThemeColor = (tagColor: TagColor, theme: Theme): boolean
...
+      tagColor !== 'notification-warning' &&
+      tagColor !== 'notification-success' &&
+      tagColor !== 'notification-error')
   );
 };
+
+export const getThemedBackgroundHoverColor = (tagColor: TagColor, themedColors: ThemedColors, theme: Theme): string => {
+  const isDark = isThemeDark(theme);
+  const colorMap: { [key in TagColor]: string } = {
+    'background-default': isDark ? themedColors.backgroundColorLighten : themedColors.backgroundColorDarken,
```
denyo, line 33: Couldn't the condition evaluated once instead of 10 times?  
Or the maybe even easier would be to interpolate the key, something like
```
const keySuffix = isDark ? 'Darken' : 'Lighten';
...
  'neutral-contrast-high': themedColors[`contrastHighColor${keySuffix}`]

```

packages/components/src/components/tag/tag-utils.ts
```diff
@@ -16,6 +17,29 @@ export const hasInvertedThemeColor = (tagColor: TagColor, theme: Theme): boolean
...
+    'background-surface': isDark
+      ? themedColors.backgroundSurfaceColorLighten
+      : themedColors.backgroundSurfaceColorDarken,
+    'neutral-contrast-high': isDark ? themedColors.contrastHighColorLighten : themedColors.contrastHighColor,
```
denyo, line 37: I think `contrastHighColor` should be `contrastHighColorDarken` or why break the pattern?

packages/components/src/styles/__snapshots__/colors.spec.ts.snap
```diff
@@ -4,54 +4,80 @@ exports[`getThemedColors() should return correct colors for theme: dark 1`] = `
...
   "infoColor": "#027FFC",
   "infoSoftColor": "#001A33",
+  "infoSoftColorDarken": "#040a11",
+  "infoSoftColorLighten": "#0f2943",
   "primaryColor": "#FBFCFF",
   "primaryColorDarken": "#beceff",
   "successColor": "#00C77E",
   "successColorDarken": "#008a57",
   "successSoftColor": "#003320",
+  "successSoftColorDarken": "#04110c",
```
denyo, line 35: Minor thing.. but why are new colors lower case when old colors are not?

packages/components/src/styles/colors.ts
```diff
@@ -33,11 +46,16 @@ const themeLight = {
   primaryColor: '#010205',
   primaryColorDarken: '#000000',
   backgroundColor: '#FFF',
+  backgroundColorDarken: '#e0e0e0',
```
denyo, line 49: I am wondering if all those new colors really belong into the theme or into the tag utils. Reusability is not given and I have the feeling this increases the chances to use the wrong colors anywhere else.

denyo:
I think something isn't right especially with the dark theme colors.

Bildschirm­foto 2023-01-26 um 07 17 28

Why is the surface color not matching the surface background anymore? (also applies to light theme)  
Dark theme yellow, green, red and blue are barely visible and I am wondering if this is fine a11y color contrast wise?

packages/components/src/components/tag/tag-styles.ts
affilnost: true, fixed it

packages/components/src/components/tag/tag-shared-utils.ts
affilnost: Removed it

packages/components/src/components/tag/tag-utils.ts
affilnost: it's the right line, `ThemedColors` is a type

affilnost: Fixed it

packages/components/src/components/tag/tag-utils.ts
affilnost: @denyo for "contrast high" color tag we use "primary" color as background, it's almost black for light theme (#010205). The reason for that is that Janick and Johannes told that the "contrastHighColor" from our design tokens is not black enough for "contrast hight tag".
So since  "contrast high" tag is already very black, we can't make it darker when hovering over it (or we can, but the user won't notice it). So we make it lighter when hovering over it  (contrastHighColor is lighter as primaryColor) - it's an exception from the rule.

packages/components/src/styles/colors.ts
affilnost: @denyo I added "TODO" comment into `generateThemesObject.ts`:
`
    // TODO: since such colors are being used only in Tag, it makes sense to move such colors into tag utils, in order to decrease the chances to use the wrong colors
`

I hope this is what you meant by "mark as todo".

packages/components/src/styles/__snapshots__/colors.spec.ts.snap
affilnost: @denyo all the colors, which are generated with 'tinycolor2', have lower case.. I added `toUpperCase()` when generating and made them all to be uppercase

packages/components/src/components/tag/tag-utils.ts
affilnost: Changed it as you described, with one exception (contrast high color)

packages/components/src/components/tag/tag-utils.ts
```diff
@@ -1,11 +1,12 @@
...
 import type { Theme } from '../../types';
-import { isThemeDark } from '../../utils/theme/isThemeDark'; // deep import needed since barrel contains MutationObserver and causes VRT to fail because of TAG_COLORS import
+import { isThemeDark } from '../../utils/theme/isThemeDark';
+import { ThemedColors } from '../../styles'; // deep import needed since barrel contains MutationObserver and causes VRT to fail because of TAG_COLORS import
 
 export const TAG_COLORS = [
   ...TAG_DISMISSIBLE_COLORS,
   'neutral-contrast-high',
-  'notification-neutral',
+  'notification-information',
```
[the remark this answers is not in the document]
affilnost: Changed it - now we support both: 'notification-neutral' and 'notification-information'

packages/components/src/components/tag/tag-styles.spec.ts
affilnost: Changed it - now we support both

packages/components/CHANGELOG.md
affilnost: Added diff

affilnost:
@denyo You're right, the background color for surface is not correct for vrt tests, I will correct it.

Regarding "a11y color contrast wise" - I also can't see the colors in dark so well.. I asked Johannes and Janick, and they told it's okay. Maybe it's a question to @o-hook (see the last comment from Bernd)

affilnost:
@denyo I updated the color in style.css, as well as surface color in files like packages/components/src/components/button/button.usage.md

affilnost:
@stephanschroeter I added scaling test to component-scaling.vrt.ts

affilnost:
@denyo Regarding "a11y color contrast wise" - we decided that we'll create an extra Issue for that, since it's a big change and we need to update the design tokens

packages/components/src/components/tag/tag-shared-utils.ts
affilnost: Done

packages/components/src/components/button/button.usage.md
```diff
@@ -85,7 +85,7 @@ easier for the user to tap due to a larger visible touch area and also allows to
 Though it's technically possible to use multiline text in Porsche Buttons, it's recommended to keep the Button label
 short and, therefore, avoid multiline Buttons.
 
-<div style="background:#F2F2F2; width:100%; margin-top: 64px; padding-top: 32px; padding-left: 42px; padding-bottom: 42px;">
+<div style="background:#EEEFF2; width:100%; margin-top: 64px; padding-top: 32px; padding-left: 42px; padding-bottom: 42px;">
```
stephanschroeter, line 88: check color

packages/components/src/components/button-pure/button-pure.usage.md
```diff
@@ -57,7 +57,7 @@ on the indicated action, for example by including active verbs such as "add", "e
 Multiple Buttons Pure can be combined to one Button Pure group that must be stacked left-aligned in order to guarantee
 scannability and legibility. It is recommended to not group more than 6 Buttons Pure in a row.
 
-<div style="background:#F2F2F2; width:100%; margin-top: 64px; padding-top: 32px; padding-left: 42px; padding-bottom: 42px;">
+<div style="background:#EEEFF2; width:100%; margin-top: 64px; padding-top: 32px; padding-left: 42px; padding-bottom: 42px;">
```
stephanschroeter, line 60: check color

packages/components/src/components/checkbox-wrapper/checkbox-wrapper.usage.md
```diff
@@ -42,7 +42,7 @@ Contrary to Radio Buttons, each Checkbox item can have its own validation messag
 
 Is used for a single option that the user can turn on or off (e.g. in case of a confirmation of terms and conditions).
 
-<div style="background:#F2F2F2; width:100%; margin-top: 64px; padding-top: 32px; padding-left: 42px; padding-bottom: 42px;">
+<div style="background:#EEEFF2; width:100%; margin-top: 64px; padding-top: 32px; padding-left: 42px; padding-bottom: 42px;">
```
stephanschroeter, line 45: check color

stephanschroeter (approved):

PR #2280, opened 2023-01-30
Issue/2122 Accordion

mhaegeleMHP:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2122/

#### Scope


packages/components/src/components/accordion/accordion-styles.ts
```diff
@@ -24,41 +27,59 @@ export const getComponentCss = (
...
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
+          fontSize: fontSizeText[s],
+          padding: compact ? '4px 0' : `${s === 'medium' ? '20px' : '12px'} 0`,
+        })),
+        '&::before': {
+          content: '""',
+          position: 'absolute',
+          borderRadius: borderRadiusSmall,
+          transition: getTransition('background-color'),
```
denyo, line 57: Only relevant for hover? So why not move it into the hover media query.

packages/components/src/components/accordion/accordion-styles.ts
```diff
@@ -73,9 +94,9 @@ export const getComponentCss = (
     icon: {
       width: fontLineHeight,
       height: fontLineHeight,
-      marginLeft: '1.5rem',
+      marginLeft: '24px',
```
denyo, line 97: Could use flex gap?

packages/components/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - `underline` prop for `Link Pure`
 - `Checkbox Wrapper`, `Radio Button Wrapper`, `Popover` and `Tag Dismissible` supports `theme` prop
 - `Icon` supports `x-small` and `x-large` values of `size` prop
+- `Accordion` `size="medium"` in `compact` variant
```
denyo, line 17: Not sure what this means.

mhaegeleMHP: Size medium was only available when `compact="false"`. I wrote it a bit different to clarify

packages/components/src/components/accordion/accordion-styles.ts
mhaegeleMHP: You want the transition on hover and when the mouse moves away from the element. If we put it into the hover media query we only have the transition on hover, not when it moves out.

denyo: No, that's the case when you put it into the `:hover` pseudo selector. The hover media query is supposed to wrap the pseudo selector and should also wrap the transition.

packages/components/src/components/accordion/accordion-styles.ts
```diff
@@ -22,43 +25,62 @@ export const getComponentCss = (
...
+        '&::before': {
+          content: '""',
+          position: 'absolute',
+          borderRadius: borderRadiusSmall,
+          transition: getTransition('background-color'),
+          ...(compact
+            ? {
+                top: '2px',
+                bottom: '2px',
+                left: '-4px',
```
denyo, line 63: Identical for both cases and can therefore be moved out of the condition.

denyo (approved):

PR #2281, opened 2023-01-30
Apply V3 Design / Modal - Issue/v3-2137

o-hook:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/v3-2137/components/modal/examples

#### Scope


packages/components/src/components/modal/modal-styles.ts
```diff
@@ -106,20 +111,29 @@ export const getComponentCss = (
...
           },
         }),
         overflowY: 'auto', // overrideable
       },
       '::slotted': addImportantToEachRule({
-        ...getSlottedJssStyle(32, hasHeader),
-        [mediaQueryM]: getSlottedJssStyle(40, hasHeader),
-        [mediaQueryXxl]: getSlottedJssStyle(64, hasHeader),
+        ...mergeDeep(
+          getSlottedJssStyle(32, hasHeader, disableCloseButton),
```
denyo, line 122: Nested object composition seems unnecessary in front of `mergeDeep`.
Also slotted jss might be confusing since this usually refers to light dom styles that are injected into the head.

packages/components/src/components/modal/modal-styles.ts
```diff
@@ -106,20 +111,29 @@ export const getComponentCss = (
...
+          buildResponsiveStyles(fullscreen, (fullscreenValue: boolean) => ({
+            [`&(.${stretchToFullModalWidthClassName}:first-child)`]: {
```
denyo, line 124: Pseudo selectors might be nested one level deeper for less redundancy.

packages/components/src/components/modal/modal-styles.ts
```diff
@@ -128,41 +142,44 @@ export const getComponentCss = (
...
-          padding: pxToRemWithUnit(40),
+        outline: 0,
+        '&:focus::before': {
+          content: '""',
+          position: 'fixed',
+          border: `${borderWidthBase} solid`,
+          ...buildResponsiveStyles(fullscreen, (fullscreenValue: boolean) => ({
+            borderRadius: fullscreenValue ? 0 : '12px',
+            borderColor: fullscreenValue ? lightThemePrimaryColor : darkThemePrimaryColor,
+            ...(fullscreenValue ? getInsetJssStyle(0) : getInsetJssStyle(-4)),
```
denyo, line 155: Could move condition next to function parameter like `getInsetJssStyle(fullscreenValue ? 0 : -4)`.

packages/components/src/components/modal/modal-styles.ts
```diff
@@ -128,41 +142,44 @@ export const getComponentCss = (
...
-      top: 0,
-      right: 0,
-      padding: pxToRemWithUnit(8),
-      border: `${pxToRemWithUnit(6)} solid ${lightThemeBackgroundColor}`,
-      background: lightThemeBackgroundColor,
+      top: '8px',
+      right: '8px',
+      border: `2px solid ${lightThemeBackgroundColor}`, // needed to enlarge button slightly without affecting the hover area (are equal now).
+      borderRadius: '4px',
+      backgroundColor: lightThemeBackgroundColor,
```
denyo, line 178: Could use `background` shorthand as it was before.

packages/components/src/components/modal/modal-styles.ts
```diff
@@ -2,23 +2,27 @@ import type { JssStyle } from 'jss';
...
   addImportantToEachRule,
   getBaseSlottedStyles,
-  getFocusJssStyle,
   getInsetJssStyle,
   getThemedColors,
   pxToRemWithUnit,
 } from '../../styles';
 import { MODAL_Z_INDEX } from '../../constants';
-import { getFocusVisibleFallback } from '../../styles/focus-visible-fallback';
+import { themeDarkBackgroundShading } from '@porsche-design-system/utilities-v2/src/js/theme/themeDarkBackgroundShading';
```
denyo, line 21: Shouldn't be imported from source but from built package instead.
Perhaps linting rules need to be adjusted?

packages/components-angular/tests/vrt/specs/modal.vrt.ts
```diff
@@ -5,12 +5,12 @@ import {
...
 } from '@porsche-design-system/shared/testing';
 
-xit.each(extendedViewports)('should have no visual regression for basic modal for viewport %s', async (viewport) => {
+it.each(extendedViewports)('should have no visual regression for basic modal for viewport %s', async (viewport) => {
   expect(await vrtTest(getVisualRegressionTester(viewport), 'modal-basic', '/modal-basic')).toBeFalsy();
 });
 
 // TODO: flaky vrt test, check in a newer version of puppeteer if issue still remains
-xit.each(extendedViewports)(
+it.each(extendedViewports)(
```
denyo, line 13: Comment above seems invalid now.

packages/components-react/tests/vrt/specs/modal.vrt.ts
```diff
@@ -5,12 +5,12 @@ import {
...
 } from '@porsche-design-system/shared/testing';
 
-xit.each(extendedViewports)('should have no visual regression for basic modal for viewport %s', async (viewport) => {
+it.each(extendedViewports)('should have no visual regression for basic modal for viewport %s', async (viewport) => {
   expect(await vrtTest(getVisualRegressionTester(viewport), 'modal-basic', '/modal-basic')).toBeFalsy();
 });
 
 // TODO: flaky vrt test, check in a newer version of puppeteer if issue still remains
-xit.each(extendedViewports)(
+it.each(extendedViewports)(
```
denyo, line 13: Invalid comment in previous line.

packages/components/src/components/modal/modal-styles.ts
o-hook: Nested object comp is fixed. But what do you mean with "confusing slotted CSS"? We need this to react to the border-radius in relation to the fullscreen mode (with no border radius).

packages/components-react/tests/vrt/specs/modal.vrt.ts
o-hook: Same as preview comment ;-)

packages/components/src/components/modal/modal-styles.ts
o-hook: `themeDarkBackgroundShading` is not included into the build atm, because it's for internal usage only.

o-hook: `themeDarkBackgroundShading` is currently not included into the build because it is for internal usage only (I guess).

packages/components/src/components/modal/modal-styles.ts
denyo: Slotted css refers to the styles that are injected into the head that look like `p-text a { ... }`.
But in this context it is about the `:slotted` selector as far as I can tell which are styles that are in the shadow dom.

o-hook: Ah, ok, but I didn't created this function name ;-) But I can rename it if you want.

o-hook: Another note: It seems that `...mergeDeep` with nested object comp is needed. Otherwise some selectors get stripped off.

packages/components/src/components/modal/modal-styles.ts
o-hook: I've added it to `utilities` package and changed the import.

denyo:
This scenario might a bit unlucky but I guess you are aware of it.
Bildschirm­foto 2023-02-06 um 09 11 10

denyo (approved):

o-hook:
>

PR #2287, opened 2023-02-02
Apply V3 Design / Grid (issue/2155)

stephanschroeter:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2155/components/grid

#### Scope


packages/components/src/components/grid/grid-item/grid-item-styles.ts
```diff
@@ -1,21 +1,23 @@
-import type { GridGutter } from '../grid/grid-utils';
 import type { GridItemOffset, GridItemSize } from './grid-item-utils';
-import { buildResponsiveStyles, getCss, mergeDeep } from '../../../utils';
-import { addImportantToEachRule, pxToRemWithUnit } from '../../../styles';
 import type { BreakpointCustomizable } from '../../../types';
+import { buildResponsiveStyles, getCss, mergeDeep } from '../../../utils';
+import { addImportantToEachRule } from '../../../styles';
+import { gridGap } from '@porsche-design-system/utilities-v2';
 
+const gutter = `calc(${gridGap} / 2)`;
```
denyo, line 7: Could be shared instead of redefined.

packages/components/src/utils/log/warnIfDeprecatedComponentIsUsed.ts
```diff
@@ -0,0 +1,10 @@
+import { getTagNameWithoutPrefix } from '../tag-name';
+
+// TODO: add missing unit test
+export const warnIfDeprecatedComponentIsUsed = (host: HTMLElement, message?: string): void => {
```
denyo, line 4: Should be combined or replaced with `warnIfComponentIsDeprecated` util.

packages/components/src/components/grid/grid-item/grid-item-styles.ts
stephanschroeter: I don't think it will save a lot since the other gutter needs to be negative.

denyo (approved):

PR #2340, opened 2023-02-17
Issue/2124 button group

tatjana-gartner2:
### Pull Request Checklist

#### References

#### Scope


packages/components/src/components/button-group/button-group-styles.ts
```diff
@@ -47,13 +27,10 @@ export const getComponentCss = (direction: BreakpointCustomizable<ButtonGroupDir
       },
       div: {
         display: 'flex',
-        marginTop: `-${spacingStaticMedium}`,
+        marginTop: `${spacingFluidSmall}`,
```
denyo, line 30: String interpolation is useless.
I think something is wrong now, because this adds a `margin-top` to the component where there wasn't any in the past.

denyo:
Some icons / wordings are missing in VRTs.
Bildschirm­foto 2023-02-20 um 10 00 15

denyo:
What's going on with the alignment here?

Bildschirm­foto 2023-02-20 um 09 59 35

packages/components/src/components/button-group/button-group-styles.ts
tatjana-gartner2: In the past the margin was removed from the button group wrapper and added to the button itself. This also increases the distance between all buttons when direction="row" is selected. I didn't see any need for the margin to be on the buttons themselves, so I removed the margin from the buttons and set it to the button group.

The spacing is bigger now because of using the fluid spacing (discussed this with johannes on friday)

stephanschroeter: I also don't think that marginTop is needed at all since, everything is now handled with flex gap.

denyo (approved):

packages/components/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Background color of `Scroller`'s `prev` and `next` buttons in dark theme
 - Partials: Removed deprecated `withoutTags` option for all partials, please use `format: 'jsx'` instead
 - `Divider` matches new design language.
+- `Button Group` matches new design language.
```
denyo, line 17: Bad merge? Should be combined with line above.

packages/components/CHANGELOG.md
```diff
@@ -405,21 +406,6 @@ to the new values since those ones will be removed with next major version.
 + …addMessage({ text: `Some message`, state: 'info' })
 ```
 
-##### Divider:
-
```
denyo, line 409: Why was this removed?

denyo (changes requested):
See comments.

denyo (approved):

PR #2375, opened 2023-02-28
Housekeeping/color filters and ssr snapshots

denyo:
### Pull Request Checklist

#### References

#### Scope


packages/components/src/styles/color-filters/toFilter.ts
```diff
@@ -0,0 +1,2 @@
+export const toFilter = (values: [number, number, number, number, number, number]): string =>
```
mhaegeleMHP, line 1: would be more readable and safer to use if the parameters are named, eg. invert: number?

stephanschroeter (approved):

packages/components/src/styles/color-filters/toFilter.ts
denyo: `values` is the parameter, the rest are types.

PR #2381, opened 2023-03-02
Issue/2361 button-tile component

tatjana-gartner2:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2361/

#### Scope


packages/components/src/components/button-tile/button-tile.tsx
```diff
@@ -0,0 +1,182 @@
...
+      </PrefixedTagNames.pButtonPure>
+    );
+
+    return (
+      <div class="root">
+        <div class="image-container">
+          <slot />
+        </div>
+        <div class="content">
+          <button {...sharedButtonProps} class="button-overlay" tabIndex={-1} aria-hidden="true"></button>
```
denyo, line 175: Could be self closing.

packages/components/src/components/button-tile/button-tile.tsx
```diff
@@ -0,0 +1,182 @@
...
+  LINK_BUTTON_TILE_ALIGNS,
+  throwIfAlignTopAndNotCompact,
+  isDisabledOrLoading,
+} from '../../utils';
+import { getComponentCss } from './button-tile-styles';
+import { getSlottedCss } from '../../styles/link-button-tile-styles';
+import { getButtonAriaAttributes } from '../button/button-utils';
+import { ButtonProps } from './button-tile-utils';
+
+const propTypes: PropTypes<typeof ButtonTile> = {
```
denyo, line 37: Hmm, now basically everything is duplicated?
I thought we wanted to abstract or reuse a thing or two.

packages/components/src/styles/link-button-tile-styles.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { getSlottedCss } from './link-button-tile-styles';
+
+describe('getSlottedCss()', () => {
```
denyo, line 3: Can be removed together with the implementation.

packages/components/src/components/button-tile/button-tile-utils.ts
```diff
@@ -0,0 +1,10 @@
+import { Theme } from '../../utils/theme';
+import { ButtonVariant, LinkButtonIconName } from '../../types';
+
+export type ButtonProps = {
+  theme: Theme;
+  variant: ButtonVariant;
+  AriaAttributes?;
```
denyo, line 7: Looks like something is wrong here.

packages/components/src/components/button-tile/button-tile-styles.ts
```diff
@@ -0,0 +1,37 @@
...
+  align: LinkButtonTileAlign,
+  compact: BreakpointCustomizable<boolean>,
+  hasGradient: boolean,
+  isDisabled: boolean,
+  isLoading: boolean
+): string => {
+  const disabledOrLoading = isDisabledOrLoading(isDisabled, isLoading);
+
+  return getCss(
+    mergeDeep(getLinkButtonTileStyles(aspectRatio, size, weight, align, compact, hasGradient, isDisabled), {
```
denyo, line 25: Why do we need a `mergeDeep` here?

packages/components/src/utils/link-button-tile/link-button-tile-weight.ts
```diff
@@ -0,0 +1,2 @@
+export const LINK_BUTTON_TILE_WEIGHTS = ['regular', 'semibold'] as const;
```
denyo, line 1: What's the benefit of having separate files for each array and type?

packages/components/src/components/button-tile/button-tile.tsx
tatjana-gartner2: Yes, you are right. I will have a look at this :)

packages/components/src/utils/link-button-tile/link-button-tile-utils.ts
```diff
@@ -0,0 +1,25 @@
+import { BreakpointCustomizable } from '../breakpoint-customizable';
```
affilnost, line 1:
```suggestion
import type { BreakpointCustomizable } from '../breakpoint-customizable';
```

packages/components/src/utils/link-button-tile/link-button-tile-utils.spec.ts
```diff
@@ -1,8 +1,8 @@
-import { LinkTileAlign, throwIfAlignTopAndNotCompact } from './link-tile-utils';
+import { LinkButtonTileAlign, throwIfAlignTopAndNotCompact } from './link-button-tile-utils';
 import { BreakpointCustomizable } from '../../utils/breakpoint-customizable';
```
affilnost, line 2:
```suggestion
import type { BreakpointCustomizable } from '../../utils/breakpoint-customizable';
```

packages/components/src/utils/link-button-tile/link-button-tile-utils.spec.ts
```diff
@@ -1,8 +1,8 @@
-import { LinkTileAlign, throwIfAlignTopAndNotCompact } from './link-tile-utils';
+import { LinkButtonTileAlign, throwIfAlignTopAndNotCompact } from './link-button-tile-utils';
```
affilnost, line 1:
```suggestion
import type { LinkButtonTileAlign } from './link-button-tile-utils';
import { throwIfAlignTopAndNotCompact } from './link-button-tile-utils';
```

packages/components/src/components/button-tile/button-tile.tsx
```diff
@@ -0,0 +1,170 @@
...
+  validateProps,
+  throwIfAlignTopAndNotCompact,
+  isDisabledOrLoading,
+  LINK_BUTTON_TILE_SIZES,
+  LINK_BUTTON_TILE_WEIGHTS,
+  LINK_BUTTON_TILE_ASPECT_RATIOS,
+  LINK_BUTTON_TILE_ALIGNS,
+} from '../../utils';
+import { getComponentCss } from './button-tile-styles';
+import { ButtonTileProps } from './button-tile-utils';
```
affilnost, line 27:
```suggestion
import type { ButtonTileProps } from './button-tile-utils';
```

packages/components/src/styles/link-button-tile-styles.ts
```diff
@@ -0,0 +1,164 @@
...
+  isDisabled?: boolean,
+  isDIsabledOrLoading?: boolean
+): Styles => {
+  const isTopAligned = align === 'top';
+  return {
+    '@global': {
+      ':host': {
+        display: 'block',
+        ...addImportantToEachRule(hostHiddenStyles),
+        cursor: isDIsabledOrLoading ? 'not-allowed' : 'pointer',
```
affilnost, line 74:
```suggestion
        cursor: isDisabledOrLoading ? 'not-allowed' : 'pointer',
```

packages/components/src/components/button-tile/button-tile-utils.ts
```diff
@@ -0,0 +1,9 @@
+import { Theme } from '../../utils/theme';
```
affilnost, line 1:
```suggestion
import type { Theme } from '../../utils/theme';
```

packages/components/src/components/button-tile/button-tile-utils.ts
```diff
@@ -0,0 +1,9 @@
...
+import { ButtonVariant, LinkButtonIconName } from '../../types';
```
affilnost, line 2:
```suggestion
import type { ButtonVariant, LinkButtonIconName } from '../../types';
```

affilnost:
"Scaling" and "High contrast" vrt-tests are missing

packages/components/CHANGELOG.md
```diff
@@ -14,7 +14,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - `Display`: value `small` for prop `size`
 - Styles: `displaySmallStyle` and `pds-display-small`
 - Styles: `fontSizeDisplaySmall` and `$pds-font-size-display-small`
-
+  -`Button Tile`
```
denyo, line 17:
```suggestion
- `Button Tile`
```

packages/components/scripts/generateDSRComponents.ts
```diff
@@ -339,7 +338,7 @@ $&`
           );
       }
 
-      return newFileContent;
+      return '// @ts-nocheck\n' + newFileContent;
```
denyo, line 341: That's not good. We don't want to disable every type checking in those generated files.  
Instead, the `implementes XY` part should be removed since the suggested change tries to solve that, right?

packages/components/src/components/button-tile/button-tile-styles.ts
```diff
@@ -0,0 +1,21 @@
...
+  align: ButtonTileAlign,
+  compact: BreakpointCustomizable<boolean>,
+  hasGradient: boolean,
+  isDisabled: boolean,
+  isLoading: boolean
+): string => {
+  const disabledOrLoading = isDisabledOrLoading(isDisabled, isLoading);
+
+  return getCss({
+    ...getLinkButtonTileStyles(aspectRatio, size, weight, align, compact, hasGradient, isDisabled, disabledOrLoading),
```
denyo, line 19: Useless object composition.  
In addition, instead of passing params 1 by 1, it might be possible to pass all of them into this function with the `...rest` operator, when the order of parameters is identical.

packages/components/src/components/button-tile/button-tile-utils.ts
```diff
@@ -0,0 +1,9 @@
+import type { Theme } from '../../utils/theme';
+import type { ButtonVariant, LinkButtonIconName } from '../../types';
+
+export type ButtonTileProps = {
```
denyo, line 4: Confusing name. which suggests that is is supposed to be used on the button tile component.

packages/components/src/components/button-tile/button-tile.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+    const buttonProps: ButtonTileProps = {
+      theme: 'dark',
+      variant: 'secondary',
+      icon: this.icon,
+      iconSource: this.iconSource,
+      ...getButtonAriaAttributes(this.disabled, this.loading, this.aria),
+    };
+
+    const sharedButtonProps = {
```
denyo, line 123: Why are these objects even separated when always used together?

packages/components/src/components/link-tile/link-tile-utils.ts
```diff
@@ -1,25 +1,27 @@
...
-export const throwIfAlignTopAndNotCompact = (
-  host: HTMLElement,
-  align: LinkTileAlign,
-  compact: BreakpointCustomizable<boolean>
-): void => {
-  if (align === 'top' && (!compact || (typeof compact === 'string' && compact === 'false'))) {
-    throw new Error(`Usage of ${getTagName(host)} is not valid. Top alignment is only possible when compact is true.`);
-  }
+export const sharedTilePropTypes:
+  | Omit<PropTypes<typeof ButtonTile>, 'type' | 'disabled' | 'loading' | 'icon' | 'iconSource' | 'aria'>
```
denyo, line 17: Would be great to have this without manual maintenance so that it automatically takes the common properties of both types. Something like  https://stackoverflow.com/questions/47375916/typescript-how-to-create-type-with-common-properties-of-two-types

packages/components/src/components/link-tile/link-tile.tsx
```diff
@@ -39,14 +27,14 @@ const propTypes: PropTypes<typeof LinkTile> = {
...
-export class LinkTile {
+export class LinkTile implements TileProps {
   @Element() public host!: HTMLElement;
 
   /** Font size of the description. */
   @Prop() public size?: BreakpointCustomizable<LinkTileSize> = 'default';
 
   /** Font weight of the description. */
-  @Prop() public weight?: BreakpointCustomizable<LinkTileWeight> = 'semibold';
+  @Prop() public weight?: BreakpointCustomizable<LinkTileWeight> = 'semi-bold';
```
denyo, line 37: I think the deprecation warning and mapping from `semibold` is missing?

packages/components/src/utils/link-button-tile/link-button-tile-utils.ts
```diff
@@ -0,0 +1,40 @@
...
+  host: HTMLElement,
+  align: LinkTileAlign | ButtonTileAlign,
+  compact: BreakpointCustomizable<boolean>
+): void => {
+  if (align === 'top' && (!compact || (typeof compact === 'string' && compact === 'false'))) {
+    throw new Error(`Usage of ${getTagName(host)} is not valid. Top alignment is only possible when compact is true.`);
+  }
+};
+
+export type TileProps = {
```
denyo, line 31: That's the common type used in `class X implements Y`?
Perhaps it should get a better name.

packages/components/scripts/generateDSRComponents.ts
```diff
@@ -28,6 +28,7 @@ const generateDSRComponents = (): void => {
 
       let newFileContent = fileContent
         .replace(/@Component\({[\s\S]+?\)\n/g, '')
+        .replace(/\simplements\s\w+/g, '')
```
denyo, line 31: This would be more precise and probably easier to read, too.
```suggestion
        .replace(/ implements [A-Za-z]+/g, '')
```

packages/component-meta/scripts/generateComponentMeta.ts
```diff
@@ -221,9 +221,31 @@ const generateComponentMeta = (): void => {
     }
 
     let [, rawPropTypes] = /const [a-z][a-zA-z]+: (?:Omit<)?PropTypes<.+?> = ({[\s\S]+?});/.exec(source) || [];
+
+    // handle shared props
+    let sharedPropsExportFile; // also needed for import paths of variables
+    if (rawPropTypes && rawPropTypes.match(/\.{3}[a-zA-Z]+,/g)) {
```
denyo, line 227: Could use optional chaining operator here.

packages/component-meta/scripts/generateComponentMeta.ts
```diff
@@ -221,9 +221,31 @@ const generateComponentMeta = (): void => {
...
+      const [, sharedPropsConstName] = rawPropTypes.match(/(shared[a-zA-z]+Types)/);
```
denyo, line 228: Why use a different regex than in the check above? Also why do we need to check first and then extract in a 2nd step? We could just try extract right from the beginning and then check if we extracted something.  
Also the scenario for multiple shared props isn't supported right now. This should easier be commented or supported.

packages/component-meta/scripts/generateComponentMeta.ts
```diff
@@ -221,9 +221,31 @@ const generateComponentMeta = (): void => {
...
+      const [, sharedPropsImportPath] = source.match(new RegExp(`${sharedPropsConstName}[\\s\\S]+?from '(.+)';`)) || [];
+      const componentFilePath = componentFiles.find((file) =>
```
denyo, line 230: This is kind of duplicated with the other location right? Why not make it reuasble?

packages/component-meta/scripts/generateComponentMeta.ts
```diff
@@ -221,9 +221,31 @@ const generateComponentMeta = (): void => {
...
+        file.match(new RegExp(`${tagName.replace(/^p-/, '/')}\\.tsx$`))
+      );
+      const sharedPropsExportFilePath = sharedPropsImportPath.match(/^\./)
+        ? path.resolve(componentFilePath, '..', sharedPropsImportPath) // relative path
+        : sharedPropsImportPath; // absolute path to other package
```
denyo, line 235: This isn't a scenario that will ever happen?

packages/component-meta/scripts/generateComponentMeta.ts
```diff
@@ -221,9 +221,31 @@ const generateComponentMeta = (): void => {
...
+
+      sharedPropsExportFile = fs.readFileSync(`${sharedPropsExportFilePath}.ts`, 'utf8');
```
denyo, line 237: Now we are reading this file which could just be reexporting what we are looking for.  
Why not require the constant directly via `require(pathtoFile)[nameOfConstant]` just like down below?

packages/component-meta/scripts/generateComponentMeta.ts
```diff
@@ -249,10 +271,22 @@ const generateComponentMeta = (): void => {
...
                 if (variable) {
-                  const [, variableImportPath] = source.match(new RegExp(`${variable}[\\s\\S]+?from '(.+)';`)) || [];
+                  let variableImportPath;
+                  if (
+                    (typeof variable === 'string' && source.includes(variable)) ||
+                    (typeof variable !== 'string' && variable.some((v) => source.includes(v)))
+                  ) {
+                    [, variableImportPath] = source.match(new RegExp(`${variable}[\\s\\S]+?from '(.+)';`));
+                  } else {
+                    [, variableImportPath] = sharedPropsExportFile.match(
```
denyo, line 281: Why would this be needed? Before we replaced a potentially spread  `...variable` 
 in `rawPropTypes` with its actual values. `propTypes` is based on it and should therefore be clean and contain everything we need.

packages/components-js/tests/unit/fixtures/stats.json
```diff
@@ -7,12 +7,12 @@
...
     "chunkShortName": "banner",
     "size": 5330,
-    "gzipSize": 2214
+    "gzipSize": 2216
   },
   {
     "chunkShortName": "button",
-    "size": 7288,
-    "gzipSize": 2901
+    "size": 7365,
```
denyo, line 14: Kind of strange that this got bigger. Are there some variables or functions shared now?

packages/components/CHANGELOG.md
```diff
@@ -300,6 +301,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Styles: `headingXXXLargeStyle` and `pds-heading-xxx-large`
 - Styles: `fontSizeHeadingXXLarge` and `$pds-font-size-heading-xx-large`
 
+#### 🤖 Property deprecations 🤖
```
denyo, line 304: Wrong place now but also redundant with what we already have above. So it can be removed.

packages/components/src/components/button-tile/button-tile-utils.ts
```diff
@@ -0,0 +1,16 @@
+import type { Theme } from '../../utils/theme';
+import type { ButtonAriaAttribute, ButtonType, ButtonVariant, LinkButtonIconName } from '../../types';
+
+export type ButtonTileAriaAttribute = ButtonAriaAttribute;
+export type ButtonTileIcon = LinkButtonIconName;
+export type ButtonTileType = ButtonType;
+
+export type ButtonProps = {
```
denyo, line 8: Hmm instead of defining this from scratch can't we have this based on what `Button` and `Button Pure` are offering? Similar to what we did with typing the shared propTypes.

packages/components/src/components/link-tile/link-tile.tsx
```diff
@@ -1,54 +1,43 @@
...
-  aria: AllowedTypes.aria<LinkAriaAttribute>(LINK_ARIA_ATTRIBUTES),
+  aria: AllowedTypes.aria<LinkTileAriaAttribute>(LINK_ARIA_ATTRIBUTES),
 };
 
 @Component({
   tag: 'p-link-tile',
   shadow: { delegatesFocus: true },
 })
-export class LinkTile {
+export class LinkTile implements Tile {
```
denyo, line 40: I think we need another name for `Tile` here.  Either `ITile` since `I` is a common prefix for interfaces or `ITileProps` or `ISharedTileProps` since props is the only content of this interface/type.

packages/component-meta/scripts/generateComponentMeta.ts
tatjana-gartner2: Correct, but but the single importpaths of the shared props can't be found in the component anymore. The single importpaths can only be found in the file, where the shared props are defined :/

packages/components-js/tests/unit/fixtures/stats.json
tatjana-gartner2: Nothing changed in the button component, the button-tile uses some of the types of the button component and also has the nested button. I am not sure why this could impact the chunk size of the button component 🙈

packages/components/src/components/button-tile/button-tile-utils.ts
tatjana-gartner2: Actually this type is not defined within the button component, so unfortunately there is nothing to reuse. The type here is only needed for react 
image

packages/component-meta/scripts/generateComponentMeta.ts
```diff
@@ -107,7 +123,7 @@ const generateComponentMeta = (): void => {
   }, {} as Record<TagName, string>);
 
   const meta: ComponentsMeta = TAG_NAMES.reduce((result, tagName) => {
-    const source = componentSourceCode[tagName];
+    let source = componentSourceCode[tagName];
```
denyo, line 126: I think we shouldn't modify the original. Instead keep it as it as and have the modified one in addition and only use it where needed.

packages/component-meta/scripts/generateComponentMeta.ts
```diff
@@ -220,10 +236,43 @@ const generateComponentMeta = (): void => {
...
-    rawPropTypes = rawPropTypes
-      ?.replace(/([a-zA-Z]+): (.+),/g, '$1: "$2",') // wrap values in quotes to make the object evaluable
-      .replace(/[^"](AllowedTypes\.(?:shape|oneOf).+\([{[][\s\S]+?[}\]]\)),/g, '`$1`,'); // wrap multiline shape object and oneOf array in backticks
+    let [, rawPropTypes] = /const [a-z][a-zA-Z]+: (?:Omit<)?PropTypes<.+?> = ({[\s\S]+?});/.exec(source) || [];
+
+    // handle shared props
+    const sharedPropsName = rawPropTypes?.match(/(?<=\.{3})(([a-zA-Z]+)(?=,))/g);
+    if (sharedPropsName) {
+      // TODO: currently the scenario for multiple shared props is not supported
+      if (sharedPropsName.length > 1) {
```
denyo, line 245: I think this isn't right since `match()` always returns an array? Might be mistaken with `matchAll()`?

packages/component-meta/scripts/generateComponentMeta.ts
```diff
@@ -220,10 +236,43 @@ const generateComponentMeta = (): void => {
...
+        throw new Error('Currently the scenario for multiple shared props is not supported.');
+      }
+      const sharedPropsConstName = sharedPropsName[0];
```
denyo, line 248: Can be destructured like this.
```suggestion
      const [sharedPropsConstName] = sharedPropsName;
```

packages/component-meta/scripts/generateComponentMeta.ts
```diff
@@ -220,10 +236,43 @@ const generateComponentMeta = (): void => {
...
+      const allSharedConstants = sharedPropsExportFile.match(/([A-Z]{2,}(_[A-Z]+)*)/g).join(', ');
+      source = `import { ${allSharedConstants} } from '../../utils';\n${source}`;
+
+      // since it is more verbose to address shared proptypes using require(),
+      // the next steps are to ensure the extracted object matches the required type
+      const evaluableSharedProps = getEvaluablePropTypeString(sharedProps);
+      const requiredKeys = Object.keys(require(sharedPropsExportFilePath)[sharedPropsConstName]).sort();
+      const readFileKeys = Object.keys(eval(`(${evaluableSharedProps})`) as Record<string, string>).sort();
+      // TODO: currently the scenario for shared props imported from multiple files is not supported
+      if (JSON.stringify(requiredKeys) !== JSON.stringify(readFileKeys)) {
```
denyo, line 266: 🤓 comparing key by key would be easier and less overhead than the json approach.

packages/component-meta/scripts/generateComponentMeta.ts
```diff
@@ -220,10 +236,43 @@ const generateComponentMeta = (): void => {
...
+        throw new Error('Currently the scenario for shared props imported from multiple files is not supported.');
+      }
+
+      sharedProps = sharedProps.replace(/([{}])/g, '');
+      rawPropTypes = rawPropTypes.replace(/\.{3}[a-zA-Z]+,/g, sharedProps);
```
denyo, line 271: How can we be sure that we are actually replacing the right sharedProps? Shouldn't we use the previously extracted string?

packages/component-meta/scripts/generateComponentMeta.ts
```diff
@@ -6,12 +6,28 @@ import { TAG_NAMES, INTERNAL_TAG_NAMES } from '@porsche-design-system/shared';
 import type { TagName } from '@porsche-design-system/shared';
 
 const glue = '\n\n';
+// can't resolve @porsche-design-system/components without building it first, therefore we use relative path
+const sourceDirectory = path.resolve('../components/src/components');
+const componentFiles = globby.sync(`${sourceDirectory}/**/*.tsx`);
+
+const getExportFilePath = (source: string, constName: string | string[], tagName: TagName): string => {
+  const [, importPath] = source.match(new RegExp(`${constName}[\\s\\S]+?from '(.+)';`));
```
denyo, line 14: A `string[]` parameter most likely messes up this regex. 
Bildschirm­foto 2023-03-16 um 09 49 59

packages/components/src/components/button-tile/button-tile-utils.ts
denyo: Not sure if we are talking about the same thing.  
This type is used for typing props that are then used on a `p-button` or `p-button-pure`.  So why do we have to define this type manually and can't have it based on what is available from `p-button` or `p-button-pure` by using `InstanceType` or something?

PR #2389, opened 2023-03-07
Housekeeping/fix ux upgrade dep

mhaegeleMHP:
### Pull Request Checklist

#### References

#### Scope


packages/components-angular/package.json
```diff
@@ -41,7 +41,7 @@
     "angular-imask": "6.4.2",
     "imask": "^6.4.3",
     "rxjs": "^7.8.0",
-    "tslib": "^2.4.1",
+    "tslib": "^2.5.0",
```
affilnost, line 44: I noticed that we have `"tslib": "^2.0.1"` in `/packages/components-angular/projects/angular-wrapper/package.json`. Should we update it there too?

packages/components-js/package.json
```diff
@@ -75,18 +75,18 @@
...
+    "devtools-protocol": "^0.0.1112051",
     "globby": "^11.0.4",
     "js-beautify": "^1.14.7",
     "jsdom": "^20.0.3",
-    "local-web-server": "^5.2.1",
+    "local-web-server": "^5.3.0",
     "terser-webpack-plugin": "^5.3.6",
     "ts-node": "^10.9.1",
-    "typescript": "~4.9.4",
+    "typescript": "~4.9.5",
```
affilnost, line 87: I noticed that we have `"typescript": "~4.6.4"` in `/packages/components-angular/package.json` - should we also update it there?

packages/components-vue/package.json
```diff
@@ -33,32 +33,32 @@
   },
   "dependencies": {
     "@porsche-design-system/components-vue": "3.0.0-alpha.2",
-    "vue": "^3.2.45",
+    "vue": "^3.2.47",
```
affilnost, line 36: I noticed that we have: `"vue": "^2.7.10",` in `/packages/storefront/package.json` and `"vue": ">=3.0.0 <4.0.0"` in `/packages/components-vue/projects/vue-wrapper/package.json` - should we also update it there?

packages/components/package.json
```diff
@@ -47,31 +47,31 @@
     "@porsche-design-system/utilities-v2": "0.0.0",
     "@splidejs/splide": "^4.1.4",
     "@stencil/core": "~2.20.0",
-    "jss": "^10.9.2",
+    "jss": "^10.10.0",
```
affilnost, line 50: I noticed that we have `"jss": "^10.9.2",` in `/packages/utilities/projects/styles/package.json` - should we also update it there?

packages/components/package.json
```diff
@@ -47,31 +47,31 @@
...
     "jss-plugin-sort-css-media-queries": "^1.0.1-beta.0",
-    "jss-preset-default": "^10.9.2",
+    "jss-preset-default": "^10.10.0",
```
affilnost, line 52: I noticed that we have `"jss-preset-default": "^10.9.2",`, in `/packages/utilities/projects/styles/package.json` - should we also update it there?

packages/components-vue/package.json
mhaegeleMHP: There is a difference between the vue package which contains the vue 3 wrappers and our storefront, which is still on vue 2. See dependencies .md file.

packages/components-js/package.json
mhaegeleMHP: Angular decides which typescript versions are allowed, depending on the version of angular we use. Thats why its pinned. See dependencies.md.

packages/components-angular/package.json
mhaegeleMHP: They are upgraded as we can see in the yarn.lock. It seems they got previously upgraded via deletion of yarn.lock thats why its not mirrored in the package.json.

packages/components/package.json
mhaegeleMHP: See comment in angular

packages/components/package.json
mhaegeleMHP: See comment in angular

denyo (approved):

Issue #2487, opened 2023-04-14
V3 Status Fixes


marcelbertram:
Scope
---
As part of the V3 migration we still have some general and component specific improvement tasks.

Notes
---
- Grab some items from the list and document in subtasks and communicate in channel to avoid double work
- As part of the issue update the status overview file and remove completed topics
- https://designsystem.porsche.com/v3/component-v3-status
- use housekeeping branch

Acceptance Criteria
---
- [x] V3 Status fixes are closed and list updated

Out of scope
---
- Accordion Optimize icon (size) for compact mode #2309 
- Accordion hover state 
- [x] Radio focus is getting lost in some circumstances (#2206)
- [x] Naming of banner width should probably be aligned with what's going on in the carousel since there is similar behavior
- [x] Accordion Remove resize-observer fallback
- [x] Pagination Safari + Safari iOS: Prev button always looks disabled
- [x] Checkbox intermediate example only shows the first checkbox as intermediate.
- [x] Use px instead of rem apart from font-size and get rid of pxToRemWithUnit() where possible
- [x] modal ::before element has visible border (2px solid) #1971 
- [x] Add letter-spacing: normal; as fontLetterSpacing and reuse in typography styles (Name)
- [x] Link-Tile model signature Panamera is overlapping the padding of on small viewports 320px (https://github.com/porsche-design-system/porsche-design-system/issues/2487#issuecomment-1512890112)
- [x] Model signature: Change color in Safari when changing the theme programmatically
- [x] Toast External @Method() addMessage() must return a Promise
- [x] Internalt Rename utilities/projects/{styles/utilities}
- [x] Pagination: Get rid of counterReset (no JS for SSR) / Pagination shouldn't loose focus by tabindex="0" when prev/next buttons are getting aria-disabled="true" #578 

Subtasks
---
- [x] Update "Usage Docs missing" entries in "V3 Status" (Anastasiia) [PR](https://github.com/porsche-design-system/porsche-design-system/pull/2497)
- [x] Carousel -> Safari iOS: align-header="center" has no effect (Anastasiia) [PR](https://github.com/porsche-design-system/porsche-design-system/pull/2499)
- ~~[ ] Text|Heading|Display: Styling could be on :host instead of .root (Tatjana) [PR](https://github.com/porsche-design-system/porsche-design-system/pull/2502)~~
- [x] Text|Heading|Display: Reuse style across component
[...]

Issue #2390, opened 2023-03-09
[BUG] - radio-button-wrapper throws error with exotic name properties


Rouby:
### Contact Details

jonathan.burke@codecentric.de

### JS Framework

React

### Porsche Design System version

2.19.0

### Reproduction Example

https://stackblitz.com/edit/react-qh8ucg?file=App.tsx

### Steps to reproduce the behavior

1. Switch radio-button selection
2. Error gets thrown in the console

### Expected behavior

No console error should be thrown.

### Relevant log output

```shell
Uncaught DOMException: Failed to execute 'querySelectorAll' on 'Document': 'input[type=radio][name=some.nested.value]' is not a valid selector.
    at HTMLInputElement.p (https://cdn.ui.porsche.com/porsche-design-system/components/porsche-design-system.radio-button-wrapper.345579dce7e2366608fd.js:1:4554)
```
```

Castlepool:
We're encountered the same problem with our radio buttons. Would love to see this fixed.

marcelbertram:
fixed and released in v3.0.0-alpha.6

PR #2508, opened 2023-04-19
fix(components): apply radio-button-wrapper workaround for safari als…

tatjana-gartner2:
…o on checkbox-wrapper | tg | #2487

### Pull Request Checklist

#### References

#### Scope


packages/components/src/utils/checkbox-radio-button-wrapper/checkbox-radio-button-wrapper-utils.ts
```diff
@@ -0,0 +1,12 @@
+export const changeHandler = ({ target }: Event): void =>
```
denyo, line 1: I think you could type `target` like this and wouldn't need to cast it twice.
```suggestion
export const changeHandler = ({ target }: Event & { target: HTMLInputElement }): void =>
```

packages/components/src/utils/index.ts
```diff
@@ -5,6 +5,7 @@ export { observeBreakpointChange, unobserveBreakpointChange } from './breakpoint
 export { getCurrentMatchingBreakpointValue } from './breakpoint-observer-utils';
 export * from './button-handling';
 export * from './button-link-pure-utils';
+export * from './checkbox-radio-button-wrapper/checkbox-radio-button-wrapper-utils';
```
denyo, line 8: Perhaps drop the folder nesting since there is just that one file in there?

packages/components/CHANGELOG.md
```diff
@@ -9,6 +9,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### [Unreleased]
 
+#### Fixed
+
+- `Checkbox Wrapper` Safari visual state change while hovering
```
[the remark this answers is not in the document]
tatjana-gartner2: Checkbox/Radio: Safari while hovering on label the state does not change visually: Using :host(:hover) &(input:checked) would solve the issue but to hover is applied to the whole host element. -> This task is also solved by this solution. It happened when you click the label of the checkbox and it was only visible when removing the pointer from the label.

packages/components/src/utils/checkbox-radio-button-wrapper-utils.spec.ts
```diff
@@ -0,0 +1,104 @@
+import { addChangeListener, changeHandler } from './checkbox-radio-button-wrapper-utils';
+
+describe('changeHandler()', () => {
+  it('should use correct selector for radio-button-wrapper and set display = none and immediately set it back to default value', () => {
```
denyo, line 4: To be precise, I think this util got nothing todo with any wrapper or component.
So the correct wording should mention `input type="radio". 
Please also adjust the checkbox test accordingly.

packages/components/src/utils/checkbox-radio-button-wrapper-utils.ts
```diff
@@ -0,0 +1,10 @@
+export const changeHandler = ({ target }: Event & { target: HTMLInputElement }): void =>
+  // workaround for Safari >= 15.5 -> attribute 'checked' is only set as property and not as attribute, this leads to rendering issues with keyboard navigation
+  document.querySelectorAll(`input[type=${target.type}][name=${target.name}]`).forEach((el: HTMLInputElement) => {
```
denyo, line 3: Aren't double quotes still missing for the value of the `name` attribute ?
That's a regression for this bug https://github.com/porsche-design-system/porsche-design-system/issues/2390 which was fixed here https://github.com/porsche-design-system/porsche-design-system/pull/2443.

packages/components/src/utils/checkbox-radio-button-wrapper-utils.spec.ts
```diff
@@ -0,0 +1,104 @@
+import { addChangeListener, changeHandler } from './checkbox-radio-button-wrapper-utils';
+
+describe('changeHandler()', () => {
+  it('should use correct selector for input type="radio and set display = none and immediately set it back to default value', () => {
```
denyo, line 4:
```suggestion
  it('should use correct selector for input type="radio" and set display = none and immediately set it back to default value', () => {
```

denyo (approved):

PR #2530, opened 2023-05-03
Fix broken aria-child behaviour and update to latest axe-core (Issue/2184)

o-hook:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2184/

#### Scope

- Added `aria` prop to `Scroller` component
- Update to latest `axe-core-npm` version
- Adapt `role` attributes of `Tabs Bar` and `Stepper Horizontal`


packages/components/src/aria-types.ts
```diff
@@ -187,4 +187,75 @@ export type AriaAttributes = {
   'aria-valuenow'?: number | undefined;
   /** Defines the human readable text alternative of aria-valuenow for a range widget. */
   'aria-valuetext'?: string | undefined;
-};
\ No newline at end of file
+  role?: AriaRole; };
```
denyo, line 190: Could add a new line before `};` for better formatting :)
Also should we add a comment before? Is this visible anywhere in the IDE?

denyo (approved):

packages/components/src/aria-types.ts
o-hook: Yes, for sure! Done :-)

PR #2574, opened 2023-05-24
Issue/2569

tatjana-gartner2:
### Pull Request Checklist

- [ ] I have read and accept the [Contributor License Agreement](https://github.com/porscheofficial/oss-docs/blob/67c5b24838a293ce7a964884e6005eb71f2b8579/CONTRIBUTOR_LICENSE_AGREEMENT.md)

#### References

#### Scope

### DIFF
--- a/packages/components-react/src/examples/ModalExampleChangeContent.tsx
+++ b/packages/components-react/src/examples/ModalExampleChangeContent.tsx
@@ -1,0 +1,47 @@
+import { useCallback, useState } from 'react';
+import { PButton, PModal, PText, PButtonGroup, PCheckboxWrapper } from '@porsche-design-system/components-react';
+
+export const ModalExampleChangeContentPage = (): JSX.Element => {
+  const [isModalOpen, setIsModalOpen] = useState<boolean>(false);
+  const onOpen = useCallback(() => {
+    setIsModalOpen(true);
+  }, []);
+  const onDismiss = useCallback(() => {
+    setIsModalOpen(false);
+  }, []);
+
+  const [someValue, setSomeValue] = useState<string>('Hello world!');
+
+  return (
+    <>
+      <PButton type="button" aria={{ 'aria-haspopup': 'dialog' }} onClick={onOpen}>
+        Open Modal
+      </PButton>
+
+      <PModal
+        id="modal-scrollable"
+        heading="Some Heading"
+        fullscreen={{ base: true, s: false }}
+        open={isModalOpen}
+        onDismiss={onDismiss}
+      >
+        <PText>Some Content</PText>
+        <div style={{ height: '70vh' }} />
+        <PText>
+          More Content <p>{someValue}</p>
+        </PText>
+        <PCheckboxWrapper label="Some label" hide-label="false">
+          <input type="checkbox" name="some-name-1" onChange={() => setSomeValue('Checkbox change')} />
+        </PCheckboxWrapper>
+        <div style={{ height: '40vh' }} />
+        <PText>Even More Content</PText>
+        <PButtonGroup className="footer">
+          <PButton>Save</PButton>
+          <PButton type="button" variant="secondary" icon="close">
+            Close
+          </PButton>
+        </PButtonGroup>
+      </PModal>
+    </>
+  );
+};
--- a/packages/components-react/tests/e2e/specs/modal.e2e.ts
+++ b/packages/components-react/tests/e2e/specs/modal.e2e.ts
@@ -1,0 +1,25 @@
+import type { Page } from 'puppeteer';
+import { goto, selectNode, waitForComponentsReady } from '../helpers';
+
+let page: Page;
+
+beforeEach(async () => (page = await browser.newPage()));
+afterEach(async () => await page.close());
+
+it('should not cause new lifecycle when nothing on the component changes', async () => {
+  await goto(page, 'modal-example-change-content');
+  expect(await waitForComponentsReady(page)).toBe(9); // p-modal, p-button (close), p-text (3x), p-checkbox-wrapper, p-button-group, p-button (save and close)
+
+  const host = await selectNode(page, 'p-modal');
+
+  await (await selectNode(page, 'p-button')).click();
+
+  expect(await host.evaluate((el) => el.shadowRoot.activeElement.className)).toBe('root');
+  expect(await host.evaluate((el) => el.scrollTop)).toBe(0);
+
+  await host.evaluate((el) => (el.scrollTop = 200));
+  await (await selectNode(page, 'p-checkbox-wrapper input[type="checkbox"]')).click();
+
+  expect(await page.evaluate(() => document.activeElement.tagName)).toBe('INPUT');
+  expect(await host.evaluate((el) => el.scrollTop)).toBe(200);
+});
--- a/packages/components/src/components/modal/modal.spec.ts
+++ b/packages/components/src/components/modal/modal.spec.ts
@@ -13,6 +13,7 @@ describe('modal', () => {
     component.host = document.createElement('p-modal');
     component.host.attachShadow({ mode: 'open' });
     component['closeBtn'] = document.createElement('button');
+    component['dialog'] = document.createElement('div');
   });
 
   describe('componentDidLoad', () => {
@@ -39,6 +40,28 @@ describe('modal', () => {
       component.componentDidLoad();
 
       expect(utilsSpy).not.toBeCalled();
+    });
+  });
+
+  describe('componentDidRender', () => {
+    it('should call focus() when modal is open and activeElement is not a child of host', () => {
+      component.open = true;
+      const spy = jest.spyOn(component['dialog'], 'focus');
+
+      component.componentDidRender();
+
+      expect(spy).toBeCalledWith();
+    });
+    it('should not call focus() when modal is not open and activeElement is a child of host', () => {
+      component.open = false;
+      const input = document.createElement('input');
+      input.type = 'checkbox';
+      component.host.appendChild(input);
+      const spy = jest.spyOn(component['dialog'], 'focus');
+
+      component.componentDidRender();
+
+      expect(spy).not.toBeCalledWith();
     });
   });
 
--- a/packages/components/src/components/modal/modal.tsx
+++ b/packages/components/src/components/modal/modal.tsx
@@ -90,11 +90,7 @@ export class Modal {
   public componentDidLoad(): void {
     // in case modal is rendered with open prop
     if (this.open) {
-      // reset scroll top to zero in case content is longer than viewport height, - some timeout is needed although it shouldn't
-      for (let i = 0; i < 4; i++) {
-        setTimeout(() => (this.host.scrollTop = 0), i * 5);
-      }
-
+      this.dialog.focus(); // needs to happen after render
       this.updateScrollLock(true);
     }
 
@@ -111,7 +107,14 @@ export class Modal {
 
   public componentDidRender(): void {
     if (this.open) {
-      this.dialog.focus(); // needs to happen after render
+      // fix scroll to top bug in react #2569 (react causes rerender of modal on useState hook within modal)
+      if (!this.host.contains(document.activeElement)) {
+        // reset scroll top to zero in case content is longer than viewport height, - some timeout is needed, although it shouldn't
+        for (let i = 0; i < 4; i++) {
+          setTimeout(() => (this.host.scrollTop = 0), i * 5);
+        }
+        this.dialog.focus(); // needs to happen after render
+      }
     }
   }
 
--- a/packages/components/src/components/modal/modal.tsx
+++ b/packages/components/src/components/modal/modal.tsx
@@ -90,7 +90,6 @@ export class Modal {
   public componentDidLoad(): void {
     // in case modal is rendered with open prop
     if (this.open) {
-      this.dialog.focus(); // needs to happen after render
       this.updateScrollLock(true);
     }
 
--- a/packages/components/src/components/modal/modal.tsx
+++ b/packages/components/src/components/modal/modal.tsx
@@ -90,6 +90,11 @@ export class Modal {
   public componentDidLoad(): void {
     // in case modal is rendered with open prop
     if (this.open) {
+      // reset scroll top to zero in case content is longer than viewport height, - some timeout is needed although it shouldn't
+      for (let i = 0; i < 4; i++) {
+        setTimeout(() => (this.host.scrollTop = 0), i * 5);
+      }
+
       this.updateScrollLock(true);
     }
 
@@ -106,10 +111,6 @@ export class Modal {
 
   public componentDidRender(): void {
     if (this.open) {
-      // reset scroll top to zero in case content is longer than viewport height, - some timeout is needed although it shouldn't
-      for (let i = 0; i < 4; i++) {
-        setTimeout(() => (this.host.scrollTop = 0), i * 5);
-      }
       this.dialog.focus(); // needs to happen after render
     }
   }
--- a/packages/components/src/components/modal/modal.tsx
+++ b/packages/components/src/components/modal/modal.tsx
@@ -105,15 +105,13 @@ export class Modal {
   }
 
   public componentDidRender(): void {
-    if (this.open) {
-      // fix scroll to top bug in react #2569 (react causes rerender of modal on useState hook within modal)
-      if (!this.host.contains(document.activeElement)) {
-        // reset scroll top to zero in case content is longer than viewport height, - some timeout is needed, although it shouldn't
-        for (let i = 0; i < 4; i++) {
-          setTimeout(() => (this.host.scrollTop = 0), i * 5);
-        }
-        this.dialog.focus(); // needs to happen after render
+    // fix scroll to top bug in react #2569 (react causes rerender of modal on useState hook within modal)
+    if (this.open && !this.host.contains(document.activeElement)) {
+      // reset scroll top to zero in case content is longer than viewport height, - some timeout is needed, although it shouldn't
+      for (let i = 0; i < 4; i++) {
+        setTimeout(() => (this.host.scrollTop = 0), i * 5);
       }
+      this.dialog.focus(); // needs to happen after render
     }
   }
 
--- a/packages/components-react/src/examples/index.ts
+++ b/packages/components-react/src/examples/index.ts
@@ -6,6 +6,7 @@ export * from './CarouselExampleJumpToSlide';
 export * from './InlineNotificationExampleActionButton';
 export * from './InlineNotificationExampleEvents';
 export * from './ModalExampleAccessibility';
+export * from './ModalExampleChangeContent';
 export * from './ScrollerExample';
 export * from './SegmentedControlExample';
 export * from './StepperHorizontalExample';
--- a/packages/components-react/src/routes.tsx
+++ b/packages/components-react/src/routes.tsx
@@ -2,6 +2,7 @@ import type { RouteProps } from 'react-router-dom';
 import * as fromPages from './pages';
 import * as fromExamples from './examples';
 import * as fromStyles from './styles';
+import { ModalExampleChangeContentPage } from './examples';
 
 export type RouteType = RouteProps & {
   name: string;
@@ -110,6 +111,11 @@ export const routes: RouteType[] = [
     name: 'Modal Example Accessibility',
     path: '/modal-example-accessibility',
     element: <fromExamples.ModalExampleAccessibilityPage />,
+  },
+  {
+    name: 'Modal Example Change Content',
+    path: '/modal-example-change-content',
+    element: <fromExamples.ModalExampleChangeContentPage />,
   },
   {
     name: 'Scroller Example',


packages/components-react/tests/e2e/specs/modal.e2e.ts
```diff
@@ -0,0 +1,26 @@
...
+beforeEach(async () => (page = await browser.newPage()));
+afterEach(async () => await page.close());
+
+it('should not set scrollTop = 0 and should not call dialog.focus() on changes within modal children', async () => {
+  await goto(page, 'modal-example-change-content');
+  expect(await waitForComponentsReady(page)).toBe(9); // p-modal, p-button (close), p-text (3x), p-checkbox-wrapper, p-button-group, p-button (save and close)
+
+  const host = await selectNode(page, 'p-modal');
+
+  expect(await host.evaluate((el) => el.shadowRoot.activeElement.className)).toBe('root');
```
henrijoss, line 15: Could use helper fn getActiveElementClassNameInShadowRoot

packages/components-react/tests/e2e/specs/modal.e2e.ts
```diff
@@ -0,0 +1,26 @@
...
+  expect(await host.evaluate((el) => el.scrollTop)).toBe(0);
+
+  await host.evaluate((el) => (el.scrollTop = 200));
+  await (await selectNode(page, 'p-checkbox-wrapper input[type="checkbox"]')).click();
+
+  const val = await selectNode(page, 'p-text p');
+  expect(await val.evaluate((el) => el.innerHTML)).toBe('Checkbox change');
+  expect(await selectNode(page, 'p-text p'));
+  expect(await page.evaluate(() => document.activeElement.tagName)).toBe('INPUT');
```
henrijoss, line 24: Could use helper fn getActiveElementTagName

tatjana-gartner2: this helper is not available in components-react

packages/components/src/components/modal/modal.tsx
```diff
@@ -105,8 +105,9 @@ export class Modal {
   }
 
   public componentDidRender(): void {
-    if (this.open) {
-      // reset scroll top to zero in case content is longer than viewport height, - some timeout is needed although it shouldn't
+    // fix scroll to top bug in react #2569 (useState hook within modal causes rerender of modal)
+    if (this.open && !this.host.contains(document.activeElement)) {
```
denyo, line 109: I think this isn't a good solution since this just fixes the sympton and not the actual cause.  
The real reason for the modals lifecycle and rerender is the new object that is passed to the `fullscreen` prop.

packages/components/CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### [Unreleased]
 
+#### Fixed
+
+- `Modal` scrollable modal does not jump to top on changes in react
```
denyo, line 14: I think this would also happen in other frameworks since the root cause is a new object instance triggering the web component's change detection which causes a new lifecycle.

packages/components-react/src/examples/ModalExampleChangeContent.tsx
```diff
@@ -0,0 +1,47 @@
+import { useCallback, useState } from 'react';
+import { PButton, PModal, PText, PButtonGroup, PCheckboxWrapper } from '@porsche-design-system/components-react';
+
+export const ModalExampleChangeContentPage = (): JSX.Element => {
```
denyo, line 4: This component should go into pages since examples imply that they are used in the storefront.

packages/components/src/utils/validation/validatePropChange.ts
```diff
@@ -0,0 +1,4 @@
+import { deepEqual } from '../jss';
+
+export const validatePropChange = (newVal, oldVal, propertyName, arrayOfRelevantPropNames): boolean =>
```
denyo, line 3: Does't this produce tons of lint warnings since typings are missing?
Also the functions name suggests it is validating something and throws an error.

packages/components/src/utils/validation/validatePropChange.spec.ts
```diff
@@ -0,0 +1,48 @@
+import { validatePropChange } from './validatePropChange';
+
+describe('validatePropChange()', () => {
+  const data: {
+    newVal: Record<string, any>;
+    oldVal: Record<string, any>;
+    propertyName: string;
```
denyo, line 7: When one parameter is called `arrayOfRelevantPropNames` why isn't this one called `propName`?

packages/components/src/utils/validation/validatePropChange.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+    arrayOfRelevantPropNames: string[];
+    result: boolean;
+  }[] = [
+    {
+      newVal: { foo: 'bar', bar: 'foo' },
+      oldVal: { foo: 'foo', bar: 'foo' },
+      propertyName: 'fullscreen',
+
```
denyo, line 15:
```suggestion
```

packages/components/src/utils/validation/validatePropChange.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+      arrayOfRelevantPropNames: ['size', 'fullscreen'],
+      result: true,
+    },
+    {
+      newVal: { foo: 'bar', bar: 'foo' },
+      oldVal: { foo: 'bar', bar: 'foo' },
+      propertyName: 'fullscreen',
+      arrayOfRelevantPropNames: ['size', 'fullscreen'],
+      result: false,
+    },
```
denyo, line 32: Case for not matching propName is missing?

packages/components/src/utils/jss.ts
```diff
@@ -145,3 +145,23 @@ export const mergeDeep = <T extends Record<string, any>>(...objects: T[]): T =>
     return prev;
   }, {} as T);
 };
+
+export const deepEqual = <T extends Record<string, any>>(object1: T, object2: T) => {
```
denyo, line 149: Please rename to a more boolean function name, also add return type.

packages/components/src/utils/jss.ts
```diff
@@ -145,3 +145,23 @@ export const mergeDeep = <T extends Record<string, any>>(...objects: T[]): T =>
...
+  const keys1 = Object.keys(object1);
+  const keys2 = Object.keys(object2);
+
+  if (keys1.length !== keys2.length) {
```
denyo, line 153: Let's improve branching. 
If this would check for `===`, then the for loop could be within the `if` block and then we add a `else`  block with the `return true`.

packages/components/src/utils/jss.spec.ts
```diff
@@ -315,6 +316,56 @@ describe('mergeDeep()', () => {
   });
 });
 
+describe('mergeDeep()', () => {
```
denyo, line 319: ??

packages/components/src/utils/jss.spec.ts
```diff
@@ -315,6 +316,56 @@ describe('mergeDeep()', () => {
...
+      JSON.stringify(obj1), // for test description
+      JSON.stringify(obj2), // for test description
+      JSON.stringify(result), // for test description
+      obj1,
+      obj2,
+      result,
+    ])
+  )(
+    `should be called with '%s' and %s and return '%s'`,
+    (_, __, ___, obj1: Record<string, any>, obj2: Record<string, any>, result: boolean) => {
```
denyo, line 363: Explicit typings shouldn't be needed.

packages/components-js/src/pages/modal-change-content.html
```diff
@@ -0,0 +1,45 @@
...
+
+      const debugElement = document.querySelector('p');
+      debugElement.innerText = 'hello world';
+
+      const modal = document.querySelector('p-modal');
+      modal.addEventListener('dismiss', () => {
+        modal.open = false;
+      });
+      document.querySelector('p-button').addEventListener('click', () => (modal.open = true));
+      document.querySelector('.input1').addEventListener('click', () => (debugElement.innerText = 'Checkbox change'));
```
denyo, line 42: I don't see how this would reproduce the behavior of the reported bug.

packages/components-js/tests/e2e/puppeteer/specs/modal.e2e.ts
```diff
@@ -466,6 +468,24 @@ describe('focus behavior', () => {
       await page.keyboard.press('Tab');
       await expectDismissButtonToBeFocused('finally');
     });
+
+    it('should not set scrollTop = 0 on changes within modal children', async () => {
+      await goto(page, 'modal-change-content');
```
denyo, line 473: I think during e2e tests the app is neither built, nor served.
Also the example pages bring no benefit to reproduce the bahvior.

packages/components/src/components/lifecycleValidation.spec.ts
```diff
@@ -22,6 +23,9 @@ const tagNamesPublicWithProps = TAG_NAMES.filter(
 const tagNamesPublicWithoutProps = TAG_NAMES.filter(
   (tagName) => !getComponentMeta(tagName).isInternal && !getComponentMeta(tagName).props
 );
+const tagNamesWithBreakpointCustomizableProps = TAG_NAMES.filter(
+  (tagName) => getComponentMeta(tagName).breakpointCustomizableProps
```
denyo, line 27: What about other props that can be objects, like `p-table-head-cell`'s `sort`, `p-scroller`'s `scrollToPosition` and all `intl` and `aria` props?

packages/components/src/components/radio-button-wrapper/radio-button-wrapper.tsx
```diff
@@ -76,6 +77,14 @@ export class RadioButtonWrapper {
     });
   }
 
+  public componentShouldUpdate(
```
denyo, line 80: Maybe change order of defined lifecycle hooks to reflect the real order. Here it is strange that `componentDidRender` is defined above but at runtime `componentShouldUpdate` happens before.

packages/components/scripts/generateConstructorMap.ts
```diff
@@ -36,6 +36,7 @@ const generateConstructorMap = (): void => {
   componentWillLoad?: () => void;
   componentWillRender?: () => void;
   disconnectedCallback?: () => void;
+  componentShouldUpdate?: (_newValue: unknown, _oldValue: unknown, propertyName) => boolean;
```
denyo, line 39: Please remove leading underscores, also type for `propertyName` is missing.

packages/components/CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Wrong validation during SSR hydration of `Link Tile` and `Select Wrapper`
   ([#2588](https://github.com/porsche-design-system/porsche-design-system/pull/2588))
+- `Modal` scrollable modal does not jump to top on changes within dialog
```
denyo, line 16: Perhaps also mention as separate bullet point that unnecessary lifecycles are prevented when prop values don't change for complex values.

packages/components/scripts/generateConstructorMap.ts
```diff
@@ -36,6 +36,7 @@ const generateConstructorMap = (): void => {
   componentWillLoad?: () => void;
   componentWillRender?: () => void;
   disconnectedCallback?: () => void;
+  componentShouldUpdate?: (newValue: unknown, oldValue: unknown, propName?: string) => void;
```
denyo, line 39: Wrong return type? 
```suggestion
  componentShouldUpdate?: (newValue: unknown, oldValue: unknown, propName?: string) => boolean;
```

packages/components/src/components/carousel/carousel.tsx
```diff
@@ -155,6 +156,10 @@ export class Carousel {
     this.observeBreakpointChange();
   }
 
+  public componentShouldUpdate(newVal: unknown, oldVal: unknown, propName: string): boolean {
+    return !isDeepEqual(newVal, oldVal) && propName !== 'activeSlideIndex'; // we need to prevent splide reinitialization via splide.refresh() when activeSlideIndex is changed from outside
```
denyo, line 160: Switching around conditions should be a little mit more performant.

packages/components/src/components/lifecycleValidation.spec.ts
```diff
@@ -136,6 +148,20 @@ it.each<TagName>(tagNamesPublicWithoutProps)('should not call validateProps() fo
   expect(spy).not.toBeCalled();
 });
 
+it.each<TagName>(tagNamesWithPropsOfTypeObject)(
+  'should call isDeepEqual() with correct parameters for %s',
+  (tagName) => {
+    const spy = jest.spyOn(isDeepEqualUtils, 'isDeepEqual');
+    const component = componentFactory(tagName);
+
+    if (component.componentShouldUpdate) {
```
denyo, line 157: What is the check for? That lifecycle should always be defined now since it is part of the rule.

packages/components/src/components/lifecycleValidation.spec.ts
```diff
@@ -136,6 +148,20 @@ it.each<TagName>(tagNamesPublicWithoutProps)('should not call validateProps() fo
   expect(spy).not.toBeCalled();
 });
 
+it.each<TagName>(tagNamesWithPropsOfTypeObject)(
+  'should call isDeepEqual() with correct parameters for %s',
```
denyo, line 152: To match other test names:
```suggestion
  'should call isDeepEqual() with correct parameters via componentShouldUpdate for %s',
```

packages/components/src/components/lifecycleValidation.spec.ts
```diff
@@ -136,6 +148,20 @@ it.each<TagName>(tagNamesPublicWithoutProps)('should not call validateProps() fo
...
+  (tagName) => {
+    const spy = jest.spyOn(isDeepEqualUtils, 'isDeepEqual');
+    const component = componentFactory(tagName);
+
+    if (component.componentShouldUpdate) {
+      component.componentShouldUpdate('newValue', 'oldValue');
+    }
+
+    expect(spy).toBeCalledWith('newValue', 'oldValue');
```
denyo, line 161: The part where the return value of `componentShouldUpdate()` should be the opposite of what `isDeepEqual()` returns is missing?

packages/components/src/components/lifecycleValidation.spec.ts
```diff
@@ -22,6 +23,17 @@ const tagNamesPublicWithProps = TAG_NAMES.filter(
 const tagNamesPublicWithoutProps = TAG_NAMES.filter(
   (tagName) => !getComponentMeta(tagName).isInternal && !getComponentMeta(tagName).props
 );
+const tagNamesWithPropsOfTypeObject = TAG_NAMES.filter((tagName) => {
+  return (
+    getComponentMeta(tagName).breakpointCustomizableProps ||
```
denyo, line 28: That would be true for an empty array `[]` which is probably not what we want.

packages/components/src/components/lifecycleValidation.spec.ts
```diff
@@ -22,6 +23,17 @@ const tagNamesPublicWithProps = TAG_NAMES.filter(
...
+    (getComponentMeta(tagName).allowedPropValues &&
+      Object.values(getComponentMeta(tagName).allowedPropValues)
+        .map((prop) => {
```
denyo, line 31: Why so complicated and so much overhead with creating a new array? The same could be achieved with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some?retiredLocale=de

packages/components/src/components/lifecycleValidation.spec.ts
```diff
@@ -22,6 +23,17 @@ const tagNamesPublicWithProps = TAG_NAMES.filter(
...
+          return typeof prop === 'object' && !Array.isArray(prop);
```
denyo, line 32: Might be worth commenting what the array check means and that we might have an issue there in the future if a prop would accept array values.

packages/components/src/components/modal/modal.spec.ts
```diff
@@ -42,6 +43,25 @@ describe('modal', () => {
...
+
+      expect(spy).toBeCalledWith();
+    });
+    it('should not focus dialog if modal is not open', () => {
+      component.open = false;
+      const spy = jest.spyOn(component['dialog'], 'focus');
+
+      component.componentDidRender();
+
+      expect(spy).not.toBeCalledWith();
```
denyo, line 61:
```suggestion
      expect(spy).not.toBeCalled();
```

packages/components/src/components/radio-button-wrapper/radio-button-wrapper.tsx
```diff
@@ -63,6 +64,10 @@ export class RadioButtonWrapper {
     this.observeAttributes(); // once initially
   }
 
+  public componentShouldUpdate(newVal: unknown, oldVal: unknown): boolean {
+    return !isDeepEqual(newVal, oldVal);
```
denyo, line 68: Perhaps we should rename this to `hasPropValueChanged` after all? Might be more natural to read than the inversion of this utility.

packages/components/src/components/scroller/scroller.tsx
```diff
@@ -111,18 +112,24 @@ export class Scroller {
...
   }
 
+  public componentShouldUpdate(
+    newVal: unknown,
+    oldVal: unknown,
+    propName: keyof InstanceType<typeof Scroller>
+  ): boolean {
+    return (
+      !isDeepEqual(newVal, oldVal) &&
+      !(propName === 'scrollToPosition' && !isScrollable(this.isNextHidden, this.isPrevHidden)) // should only update if scrollable
```
denyo, line 122: Switching conditions around should be a little bit more performant.
In terms of lifecycle hook order it was better before since `componentDidLoad` happens first?

packages/components/src/components/toast/toast/toast.tsx
```diff
@@ -30,21 +30,17 @@ export class Toast {
     toastManager.register(this.host, () => this.toastItemElement.classList.add(toastCloseClassName));
   }
 
+  public componentShouldUpdate(_: unknown, __: unknown, propName: keyof InstanceType<typeof Toast>): boolean {
```
denyo, line 33: Order was better before.

packages/components/src/components/toast/toast/toast.spec.ts
```diff
@@ -24,18 +24,6 @@ describe('componentDidLoad', () => {
   });
 });
 
-describe('componentShouldUpdate', () => {
```
denyo, line 27: Why remove? I think this part isn't covered anywhere now.

packages/components/src/utils/is-deep-equal.ts
```diff
@@ -0,0 +1,32 @@
+import { isObject } from './jss';
+
+export const isDeepEqual = (newVal: unknown, oldVal: unknown): boolean => {
+  if (typeof newVal !== 'object' || typeof oldVal !== 'object') {
+    // string, boolean, number
+    return newVal === oldVal;
+  } else if (Array.isArray(newVal) || Array.isArray(oldVal)) {
+    return (
+      Array.isArray(newVal) &&
```
denyo, line 9: That's redundant with what has been checked in the `else if` condition already?

packages/components/src/utils/is-deep-equal.ts
```diff
@@ -0,0 +1,32 @@
...
+      Array.isArray(oldVal) &&
+      newVal.length === oldVal.length &&
+      newVal.every((item) => oldVal.includes(item)) &&
+      oldVal.every((item) => newVal.includes(item))
```
denyo, line 13: That's redundant with the previous one. Or how can it be that one array contains everything from another but not the other way around?

packages/components/src/utils/is-deep-equal.ts
```diff
@@ -0,0 +1,32 @@
...
+    );
+  }
+
```
denyo, line 16: Let's add an `else` block here for more clarity.

packages/components/src/utils/is-deep-equal.ts
tatjana-gartner2: Yes it is, but unfortunately it seems to be needed within the return again 
image

packages/components/src/utils/has-prop-value-changed.ts
```diff
@@ -0,0 +1,29 @@
+import { isObject } from './jss';
+
+export const hasPropValueChanged = (newVal: unknown, oldVal: unknown): boolean => {
+  if (typeof newVal !== 'object' || typeof oldVal !== 'object') {
+    // primitive types
+    return newVal === oldVal;
+  } else if (Array.isArray(newVal) && Array.isArray(oldVal)) {
+    // type array
+    return newVal.length === oldVal.length && newVal.every((item) => oldVal.includes(item));
```
denyo, line 9: No need to wrap in another function 
```suggestion
    return newVal.length === oldVal.length && newVal.every(oldVal.includes);
```

packages/components/src/components/lifecycleValidation.spec.ts
```diff
@@ -136,6 +148,34 @@ it.each<TagName>(tagNamesPublicWithoutProps)('should not call validateProps() fo
   expect(spy).not.toBeCalled();
 });
 
+describe.each<TagName>(tagNamesWithPropsOfTypeObject)('%s', (tagName) => {
+  const component = componentFactory(tagName);
+
+  it('should call hasPropValueChanged() with correct parameters via componentShouldUpdate', () => {
+    const spy = jest.spyOn(hasPropValueChangedUtils, 'hasPropValueChanged');
+
+    component.componentShouldUpdate('newValue', 'oldValue');
```
denyo, line 157: Perhaps add 3rd parameter since this is how the function is always called at actual runtime.
```suggestion
    component.componentShouldUpdate('newValue', 'oldValue', 'propOrStateName');
```

packages/components/scripts/generateConstructorMap.ts
```diff
@@ -36,6 +36,7 @@ const generateConstructorMap = (): void => {
   componentWillLoad?: () => void;
   componentWillRender?: () => void;
   disconnectedCallback?: () => void;
+  componentShouldUpdate?: (newValue: unknown, oldValue: unknown, propName?: string) => boolean;
```
denyo, line 39:
```suggestion
  componentShouldUpdate?: (newValue: unknown, oldValue: unknown, propName: string) => boolean;
```

denyo (approved):

PR #2627, opened 2023-06-26
Consistent and good tsconfig.json targets everywhere (Issue/2274)

### LINKED ISSUE #2274: Consistent and good tsconfig.json targets everywhere
Scope
---
We should update and align our tsconfig.json targets for both internal packages or prebuilds but also for production builds that are shipped via npm package or CDN.

Currently it is a wild mix of different targets and therefore different ES versions. This has an effect of which JS features are bundled into the build by typescript since they aren't implemented natively in the browser.  Obviously this can cause an additional overhead in the code being shipped which isn't needed.

It is also possible to go for the latest version and specify libs (e.g. from core-js) that should still be bundled.

Notes  
---
- https://www.typescriptlang.org/tsconfig#target
- https://www.typescriptlang.org/tsconfig#lib
- maybe go cutting edge internally and with npm packages since they are rebundled anyways but be more conservative with CDN build
- https://kangax.github.io/compat-table/es2016plus/
- https://caniuse.com/?search=ECMAScript (scroll to summary at very bottom and click on show more)
- Evaluate which browser supports which JS feature compared to ES version
- Evaluate if `.browserlist` can influence `tsconfig` target
- Maybe related: #1240

Acceptance criteria
---
- [x] All `tsconfig.json` targets for internal, prebuild and production builds are aligned, using a version which fits our browser support rules

Subtasks
---
- [x] Task

denyo:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2274/

#### Scope

Resolves #2274


henrijoss (approved):
Looks good to me

Issue #2577, opened 2023-05-24
[BUG] - Navigation between 2 pages is not working as expected with Stepper Horizontal states


tatjana-gartner2:
### JS Framework

angular

### Porsche Design System version

3.0.0

### What browsers are you seeing the problem on?

Chrome

### Reproduction Example

https://porschedev.slack.com/archives/CT7AVHTTQ/p1684917028128699?thread_ts=1684840923.554569&cid=CT7AVHTTQ

### Steps to reproduce the behavior

1. Go to https://stackblitz.com/edit/angular-7arksq?devToolsHeight=33&file=src%2Fstart.component.ts
2. Click on 'to second page'
3. See error

### Expected behavior

No error for the intended outcome - either by modification of the component or different implementation (docs, validation, feedback in channel)

### Relevant log output

```shell
zone.js:141 Uncaught TypeError: Cannot read properties of undefined (reading 'shadowRoot')
    at porsche-design-system.stepper-horizontal.eb1d0a47b8f56bcb98f7.js:1:11248
    at u (porsche-design-system.stepper-horizontal.eb1d0a47b8f56bcb98f7.js:1:11288)
    at z.scrollIntoView (porsche-design-system.stepper-horizontal.eb1d0a47b8f56bcb98f7.js:1:3874)
    at porsche-design-system.stepper-horizontal.eb1d0a47b8f56bcb98f7.js:1:4420
    at porsche-design-system.v3.0.0.7acdc00ea94aef5759a7.js:1:43221
    at Array.forEach (<anonymous>)
    at MutationObserver.sn (porsche-design-system.v3.0.0.7acdc00ea94aef5759a7.js:1:43200)
    at _ZoneDelegate.invoke (zone.js:368:26)
    at Zone.runGuarded (zone.js:137:47)
    at MutationObserver.<anonymous> (zone.js:121:29)
```

### Notes
- Different implementation without an error: https://stackblitz.com/edit/angular-z5tovf?devToolsHeight=33&file=src%2Fstart.component.ts,src%2Futils.ts,src%2Fapp.component.ts

DelgadoRemiz:
Hi PDS team,
are there any plans to work on this bug in next time?

PR #2641, opened 2023-07-04
[BUG] - Navigation between 2 pages is not working as expected with Stepper Horizontal states (issue/2577)

henrijoss:
…ess of child components for scrolling | hj | #2577

### Pull Request Checklist

- [x] I have read and accept the

#### References

- Preview: https://designsystem.porsche.com/issue/2577

#### Scope

#### Resolved Issue

Resolves [#2577](https://github.com/porsche-design-system/porsche-design-system/issues/2577)


packages/components/src/components/stepper-horizontal/stepper-horizontal/stepper-horizontal.tsx
```diff
@@ -102,14 +91,17 @@ export class StepperHorizontal {
     }
   }
 
+  public componentDidRender(): void {
```
denyo, line 94: We should do this in `componentDidLoad` like in every other component and not do it on every rerender.  
Are you sure you tried that lifecycle and not `componentWillLoad` by accident?

packages/components/src/components/stepper-horizontal/stepper-horizontal/stepper-horizontal.tsx
```diff
@@ -152,13 +144,16 @@ export class StepperHorizontal {
...
     this.stepperHorizontalItems = Array.from(this.host.children) as HTMLPStepperHorizontalItemElement[];
   };
 
-  private validateComponent = (): void => {
+  private validateComponentChildren = (): void => {
     throwIfChildrenAreNotOfKind(this.host, 'p-stepper-horizontal-item');
     throwIfChildCountIsExceeded(this.host, 9);
+  };
+
+  private validateComponentState = (): void => {
```
denyo, line 156: Can we combine validations? This wouldn't only make it cleaner but also more precise because when children change their might be potential 2nd current state in the new one?

denyo:
Shouldn't we add an e2e test to verify the issue is resolved?

packages/components/src/components/stepper-horizontal/stepper-horizontal/stepper-horizontal.tsx
```diff
@@ -57,25 +56,11 @@ export class StepperHorizontal {
   private currentStepIndex: number;
 
   public connectedCallback(): void {
+    this.validateComponent();
```
denyo, line 59: This is most likely to early in framework scenarios, especially Angular where properties are not bound, yet on first render. So this only reliably covers the reconnect scenario.
For the first render it would necessary to do this in `componentWillLoad` as it was before.
```suggestion
    this.validateComponent(); // on every reconnect
```

packages/components/src/components/stepper-horizontal/stepper-horizontal/stepper-horizontal.tsx
```diff
@@ -133,33 +118,33 @@ export class StepperHorizontal {
...
-    // TODO: validation? this could be any kind of dom node
-    this.stepperHorizontalItems = Array.from(this.host.children) as HTMLPStepperHorizontalItemElement[];
+    this.stepperHorizontalItems = this.createStepperHorizontalItemElements();
   };
 
   private validateComponent = (): void => {
     throwIfChildrenAreNotOfKind(this.host, 'p-stepper-horizontal-item');
     throwIfChildCountIsExceeded(this.host, 9);
-    throwIfMultipleCurrentStates(this.host, this.stepperHorizontalItems);
+    throwIfMultipleCurrentStates(this.host, this.createStepperHorizontalItemElements());
```
denyo, line 147: This got a little bit confusing and tough to minify since those class functions can't be minified. 
Why don't we drop those 1 line utils within the class and instead do this
```
this.stepperHorizontalItems = this.createStepperHorizontalItemElements();
```
as part of `validateComponent()`. 

That would be cleaner than doing those 2 steps after each other multiple times
```
this.validateComponent();
this.defineStepperHorizontalItemElements();
```

denyo (approved):

PR #2644, opened 2023-07-05
Issue/2636 - Improve bootstrapping behaviour of Icon component

tatjana-gartner2:
### Pull Request Checklist

- [ ] I have read and accept the

#### References

#### Scope

#### Resolved Issue

Resolves [IssueNumber]


henrijoss (commented):
looks good to me

packages/components/src/components/icon/icon.tsx
```diff
@@ -89,6 +89,14 @@ export class Icon {
     );
 
     // TODO: wouldn't it be better to set alt attribute instead of aria-label?
-    return <img src={buildIconUrl(this.source || this.name)} {...parseAndGetAriaAttributes(this.aria)} alt="" />;
+    return (
+      <img
+        src={buildIconUrl(this.source || this.name)}
+        {...parseAndGetAriaAttributes(this.aria)}
+        {...(!this.source && { width: 24, height: 24 })} // improve bootstrapping behaviour
```
denyo, line 96: Why conditional?

tatjana-gartner2: We should maybe not add the width and height attributes for custom icons 🤔

packages/components/src/components/icon/icon.tsx
```diff
@@ -89,6 +89,15 @@ export class Icon {
...
 
     // TODO: wouldn't it be better to set alt attribute instead of aria-label?
-    return <img src={buildIconUrl(this.source || this.name)} {...parseAndGetAriaAttributes(this.aria)} alt="" />;
+    return (
+      <img
+        src={buildIconUrl(this.source || this.name)}
+        {...parseAndGetAriaAttributes(this.aria)}
+        width={24} // improve bootstrapping behaviour
+        height={24} // improve bootstrapping behaviour
+        loading="lazy"
```
stephanschroeter, line 98: Wouldn't it be better if we'd make this configurable by a prop?

packages/components/src/components/icon/icon.tsx
```diff
@@ -89,6 +89,15 @@ export class Icon {
     );
 
     // TODO: wouldn't it be better to set alt attribute instead of aria-label?
-    return <img src={buildIconUrl(this.source || this.name)} {...parseAndGetAriaAttributes(this.aria)} alt="" />;
+    return (
+      <img
+        src={buildIconUrl(this.source || this.name)}
+        {...parseAndGetAriaAttributes(this.aria)}
+        width={24} // improve bootstrapping behaviour
```
stephanschroeter, line 96: Is this valid for custom icons passed by `source` prop too?

packages/components/src/components/icon/icon.tsx
```diff
@@ -89,6 +89,15 @@ export class Icon {
     );
 
     // TODO: wouldn't it be better to set alt attribute instead of aria-label?
-    return <img src={buildIconUrl(this.source || this.name)} {...parseAndGetAriaAttributes(this.aria)} alt="" />;
+    return (
+      <img
+        src={buildIconUrl(this.source || this.name)}
+        {...parseAndGetAriaAttributes(this.aria)}
```
stephanschroeter, line 95: Instead of using aria-label, wouldn't the `alt` attribute good enough or even better?

packages/components/src/components/icon/icon.tsx
tatjana-gartner2: As discussed in private chat, we should add the attributes also for custom sources

packages/components/src/components/icon/icon.tsx
tatjana-gartner2: As discussed in private chat, we should add the attributes also for custom sources

packages/components/src/components/icon/icon.tsx
tatjana-gartner2: As discussed in private chat, we always want to add the native prop as there is no disadvantage

denyo (approved):

Issue #2638, opened 2023-07-04
[BUG] - Contents get cropped on Carousel items


o-hook:
### Bug description
The `Carousel` components sets a border-radius on each `Carousel Item` by default. To render this border-radius, `overflow:hiddden` is set on each item which prevents contents from overflowing the `Carousel Item`.

### Contact Details

Slack Discussion: https://porschedev.slack.com/archives/CT7AVHTTQ/p1688468682936859

### JS Framework

All

### Porsche Design System version

latest

### What browsers are you seeing the problem on?

all

### Browser version

all

### Reproduction Example

- Stackblitz: https://stackblitz.com/edit/xkaf6g?file=index.html

### Expected behavior

- Contents within a `Carousel Item` are not getting cropped
- `border-radius` of `Carousel Item` is configurable/overwritable by CSS from outside

PR #2645, opened 2023-07-05
[BUG] - Contents get cropped on Carousel items (Issue/2638)

o-hook:
### Pull Request Checklist

- [x] I have read and accept the

#### References

- Preview: https://designsystem.porsche.com/issue/2638/

#### Scope

Fixed overflow hidden issue on carousel slide items and added a custom CSS property to change border-radius from outside.

#### Resolved Issue

Resolves: #2638


packages/components/src/components/carousel/carousel.examples.md
```diff
@@ -154,6 +154,13 @@ type CarouselInternationalization = {
 
 <Playground :markup="focusBehavior" :config="config"></Playground>
 
+## Custom Border Radius
+
+By default, every carousel item gets a border radius of `12px`. This can be changed via a custom CSS property, e.g.:
```
stephanschroeter, line 159:
```suggestion
By default, every carousel item gets a predefined border radius. This can be changed via a custom CSS property, e.g.:  
```

packages/components/CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - `Tabs` and `Tabs Bar` support SSR ([#2611](https://github.com/porsche-design-system/porsche-design-system/pull/2611))
 - Contents of `Tag` component can now be wrapped in multiple lines
   ([#2625](https://github.com/porsche-design-system/porsche-design-system/pull/2625))
+- `Carousel`: possibility to set custom border-radius of slide items
```
stephanschroeter, line 22:
```suggestion
- `Carousel`: Possibility to set custom border-radius of slide items
```

stephanschroeter (approved):

packages/components/src/components/carousel/carousel-utils.ts
```diff
@@ -65,11 +65,13 @@ export const warnIfHeadingIsMissing = (host: HTMLElement, heading: string): void
...
 };
 
-export const getSlidesAndAddNamedSlots = (host: HTMLElement): HTMLElement[] => {
+export const getSlidesAndAddAttributes = (host: HTMLElement): HTMLElement[] => {
   const slides = Array.from(host.children).filter(
     ({ slot }) => slot !== 'heading' && slot !== 'description'
   ) as HTMLElement[];
-  slides.forEach((el, i) => el.setAttribute('slot', `slide-${i}`));
+  slides.forEach((el, i) => {
+    Object.assign(el, { slot: `slide-${i}`, tabIndex: 0 });
```
denyo, line 73: Does this work cross browser since it implies properties are synced to attributes?

packages/components/src/components/carousel/carousel.examples.md
```diff
@@ -154,6 +154,13 @@ type CarouselInternationalization = {
 
 <Playground :markup="focusBehavior" :config="config"></Playground>
 
+## Custom Border Radius
+
+By default, every carousel item gets a border radius of `12px`. This can be changed via a custom CSS property, e.g.:  
+`--p-carousel-border-radius: 4px`.
```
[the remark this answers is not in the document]
denyo: And not hardcode 4px :)

packages/components/src/components/carousel/carousel-utils.spec.ts
```diff
@@ -114,17 +114,18 @@ describe('getSlidesAndAddNamedSlots()', () => {
...
+  it('should add incremental slot="slide-x" and tabindex="0" attributes on each child', () => {
     const host = document.createElement('p-carousel');
     const children = getChildren();
     host.append(...children);
 
-    const result = getSlidesAndAddNamedSlots(host);
+    const result = getSlidesAndAddAttributes(host);
     result.forEach((child, i) => {
       expect(child.slot).toBe(`slide-${i}`);
+      expect(child.tabIndex).toBe(0);
```
denyo, line 128: Test name doesn't match what is happening here since here we check properties, not attributes.

packages/components/src/components/carousel/carousel.examples.md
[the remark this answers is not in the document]
o-hook: Added select dropdown with border-radius values and "zero" value.

packages/components/src/components/carousel/carousel-utils.ts
o-hook: Hm, good point. So we should do it better like this?

```slides.forEach((el, i) => {
    el.setAttribute('slot', `slide-${i}`);
    el.setAttribute('tabindex', '0');
  });```

denyo: That would be the safer option, yes. But unit tests should also verify this behavior because now they check that the properties are set which might happen in jsdom but not in real browser.

PR #2676, opened 2023-07-18
Automatically detect China CDN switch (Issue/2655)

### LINKED ISSUE #2655: Automatically detect China CDN switch
Scope
---
Currently it's necessary to create two dedicated builds for a Porsche Design System component consuming team to have best CDN performance in China by defining the global variable `PORSCHE_DESIGN_SYSTEM_CDN='auto|cn'`. Unfortunately some teams don't even know about that option and it's hard for them to test the impact locally in China. Therefore, we'd like to simplify the usage by auto detect from which CDN the Porsche Design System components shall be loaded based on the domain. In case a web application is running on a `.cn` domain, then the Porsche Design System detects this automatically and loads any Porsche Design System related chunk from `cdn.ui.porsche.cn`. In any other case its using `cdn.ui.porsche.com` to fetch Porsche Design System related chunks.

Out of Scope
---
- Best CDN option for "Partial" usage can't be auto detected since the decision needs to be made during build time

Notes  
---
- https://designsystem.porsche.com/v3/must-know/performance/cdn#china-cdn
- `PORSCHE_DESIGN_SYSTEM_CDN` has no effect anymore (or it still has and is used for defining the auto detection in case the variable is undefined, then also older PDS versions would benefit)
- We should think of extending the options to `auto|cn|com`:
  - `auto` or `undefined`?: PDS decides if `com` or `cn` CDN shall be used
  - `cn`: Integrating team decides to use `cn` CDN only
  - `com` or `default`: Integrating team decides to use `com` or `default` CDN only 
- In case values are changed, keep in mind to at least create issues to update the CDN option of the Partials too (=> deprecate `auto` in favour of `com` or `default` for Partial usage).

Acceptance criteria
---
- [x] The Porsche Design System is optionally but by default able to auto detect which CDN (`cdn.ui.porsche.com` or `cdn.ui.porsche.cn`) is the best option to load its chunks + assets from during bootstrapping phase

denyo:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2655/

#### Scope

#### Resolved Issue

Resolves #2655


tatjana-gartner2 (commented):
looks good to me :)

packages/components-js/projects/components-manager-cli/library-entry/with-prefix.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+    load({ prefix: 'my-prefix' });
+    expect(spy).toBeCalledWith({ prefix: 'my-prefix', ...global.CM_CONFIG });
+  });
+
+  it('should by default set window.PORSCHE_DESIGN_SYSTEM_CDN="auto"', () => {
+    load();
+    expect(window.PORSCHE_DESIGN_SYSTEM_CDN).toBe('auto');
+  });
+
+  it('should for { cdn: "auto" } set window.PORSCHE_DESIGN_SYSTEM_CDN="auto"', () => {
```
stephanschroeter, line 41:
```suggestion
  it('should for { cdn: "auto" } set window.PORSCHE_DESIGN_SYSTEM_CDN="auto" for backwards compatiblity', () => {
```
Maybe to make things clearer?

packages/components-js/projects/components-manager-core/package.json
```diff
@@ -4,20 +4,20 @@
   "private": true,
   "license": "UNLICENSED",
   "main": "dist/index.js",
+  "module": "dist/esm/index.js",
   "types": "dist/types/index.d.ts",
   "scripts": {
-    "build": "yarn clean:build && tsc --build tsconfig.json",
+    "build": "yarn clean:build && tsc --build tsconfig.json && tsc --project tsconfig.json --module CommonJS --outDir ./dist",
```
stephanschroeter, line 10: Why is the CJS build needed? For testing reasons?

packages/components-js/projects/components-manager-core/src/services/library-handler.ts
```diff
@@ -26,15 +26,15 @@ export type LoadComponentLibraryOptions = {
...
   const { isInjected, prefixes = [], registerCustomElements } = data;
 
   const [collidingVersion] = Object.entries(getComponentsManagerData()).filter(
-    ([v, cmData]) => v !== version && cmData.prefixes.includes(prefix)
+    ([v, cmData]) => v !== version && typeof cmData !== 'string' && cmData.prefixes.includes(prefix)
   );
   if (collidingVersion) {
     throw new Error(
-      `[Porsche Design System] prefix '${prefix}' is already registered with version '${collidingVersion[0]}' of the Porsche Design System. Please use a different one.
+      `[Porsche Design System v${version}] prefix '${prefix}' is already registered with version '${collidingVersion[0]}' of the Porsche Design System. Please use a different one.
```
stephanschroeter, line 37:
```suggestion
      `[Porsche Design System] v${version} with prefix '${prefix}' is already registered with version '${collidingVersion[0]}' of the Porsche Design System. Please use a different one.
```

Not sure, if it's good to have the version within the `[]` because then we have another inconsistency within the log output compared to other logs.

packages/components-js/projects/jsdom-polyfill/rollup.config.js
```diff
@@ -24,23 +24,19 @@ export default [
...
         find: /appGlobals\.globalScripts\(\);/,
         replace: (match) => `if(!window.PDS_SKIP_FETCH) { ${match} }`,
       }),
       modify({
-        // icon svgs
-        find: /(const pdsFetch = \(input, init\) =>) (fetch\(input, init\);)/,
-        replace: (_, $1, $2) => `${$1} window.PDS_SKIP_FETCH ? Promise.resolve({ ok:true, text: () => ''}) : ${$2}`,
+        // icon svgs (img src)
+        find: /(src:) (buildIconUrl\()/,
+        replace: (_, $1, $2) => `${$1} window.PDS_SKIP_FETCH ? undefined : ${$2}`,
```
stephanschroeter, line 36: The global variable `PDS_SKIP_FETCH` is still needed, right?

packages/components-js/projects/jsdom-polyfill/src/index.js
```diff
@@ -1,34 +1,33 @@
...
 require('construct-style-sheets-polyfill');
 require('intersection-observer');
 require('matchmedia-polyfill');
 require('matchmedia-polyfill/matchMedia.addListener');
 require('scroll-behavior-polyfill');
-require('whatwg-fetch');
 
 if (!navigator.userAgent.includes('Node.js') && !navigator.userAgent.includes('jsdom')) {
-  throw new Error('This package should only be used in node and jsdom environments');
+  throw new Error('[Porsche Design System] this package should only be used in node and jsdom environments');
```
stephanschroeter, line 9:
```suggestion
  throw new Error('[Porsche Design System] the jsdom-polyfill sub package should only be used in node and jsdom environments');
```

packages/components-js/projects/components-manager-core/src/services/library-handler.ts
denyo: We usually always have the version in the brackets so it is possible to find out which prefix and team is causing it.
Bildschirmfoto 2023-08-02 um 10 41 25

packages/components-js/projects/jsdom-polyfill/rollup.config.js
denyo: Yes, but maybe not fully working everywhere since newer components were not adjusted.

packages/components-js/projects/components-manager-core/package.json
denyo: I think for jest unit tests.

stephanschroeter (approved):

PR #2677, opened 2023-07-19
[BUG] - PSelectWrapper direction "auto" fails on initial open (Issue/2620)

### LINKED ISSUE #2620: [BUG] - PSelectWrapper direction "auto" fails on initial open
### Contact Details

_No response_

### JS Framework

React

3.2.0

_No response_

### Browser version

_No response_

### Reproduction Example

https://stackblitz.com/edit/react-wq4xsn?file=App.tsx

### Steps to reproduce the behaviour

1. Open StackBlitz example
2. Click on select, adjust view-height so it should not fit on bottom
3. See that it appears on bottom, although there is not enough space

### Expected behaviour

The `Select Wrapper` should determine the position / direction automatically.

### Notes
- Why isn't an automated test not covering this scenario?
- Would it be possible to reuse the positioning mechanism of Popover?

### Relevant log output

_No response_

o-hook:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2620/

#### Scope

#### Resolved Issue

Resolves #2620


packages/components/src/components/select-wrapper/select-wrapper-dropdown/select-wrapper-dropdown-utils.ts
```diff
@@ -156,6 +154,12 @@ export const getHighlightedOptionMap = (arr: OptionMap[]): OptionMap => arr.find
 export const getValidOptions = (options: OptionMap[]): OptionMap[] =>
   options.filter((item) => !item.hidden && !item.initiallyHidden && !item.disabled);
 
+export const getLengthOfVisibleOptionsAndOptgroups = (options: OptionMap[]): number => {
+  const visibleItems = options.filter((item) => !item.hidden && !item.initiallyHidden).length;
+  const itemsWithOptgroups = options.filter((item) => item.title).length;
```
denyo, line 159: Instead of filtering and iterating over all options twice, the filter conditions should be combined.  
Also with the current implementation an option is counted two times when it is not hidden and has a title?
Perhaps also rename function to something like `getAmount..`.

packages/components/src/components/select-wrapper/select-wrapper-dropdown/select-wrapper-dropdown-utils.spec.ts
```diff
@@ -339,6 +340,21 @@ describe('getValidOptions()', () => {
   });
 });
 
+describe('getLengthOfVisibleOptionsAndOptgroups()', () => {
+  it('should return number of options that are not hidden or initiallyHidden summed with number of optgroups', () => {
```
denyo, line 344: I think we need a couple of more test cases for other scenarios like the one mentioned in the other comment.

packages/components/src/components/select-wrapper/select-wrapper-dropdown/select-wrapper-dropdown-utils.ts
o-hook: "Instead of filtering and iterating over all options twice, the filter conditions should be combined."
--> yes, I tried it with `reduce` but didn't get the requested results. Any idea how to simplify it?

"Also with the current implementation an option is counted two times when it is not hidden and has a title?"
--> this is intended, because every "title" should render an additional `optgroup`-like element which must be also taken into account of the height calculation.

o-hook: Found a solution with `reduce`. Seems to be more solid.

packages/components/src/components/select-wrapper/select-wrapper-dropdown/select-wrapper-dropdown-utils.ts
```diff
@@ -156,6 +154,14 @@ export const getHighlightedOptionMap = (arr: OptionMap[]): OptionMap => arr.find
 export const getValidOptions = (options: OptionMap[]): OptionMap[] =>
   options.filter((item) => !item.hidden && !item.initiallyHidden && !item.disabled);
 
+export const getAmountOfVisibleOptionsAndOptgroups = (options: OptionMap[]): number => {
+  return options.reduce((count, { hidden, initiallyHidden, title }) => {
```
henrijoss, line 158: Why reduce instead of filter?

o-hook: I had used `filter`before, but it worked only with two iterations, like
```
const visibleItems = options.filter((item) => !item.hidden && !item.initiallyHidden).length;
const itemsWithOptgroups = options.filter((item) => item.title).length;
return visibleItems + itemsWithOptgroups;
  ```
Couldn't find a solution how to combine it in one iteration....

henrijoss (approved):

packages/components/src/components/select-wrapper/select-wrapper-dropdown/select-wrapper-dropdown-utils.ts
o-hook: No, because we have to find both, all elements which are not "hidden" and additionally all elements with "title".

PR #2713, opened 2023-08-03
TabsBar: Incorrect semantics and keyboard behavior if slotted anchor tags are used (Issue/2040)

### LINKED ISSUE #2040: TabsBar: Incorrect semantics and keyboard behavior if slotted anchor tags are used
Scope
---
If Tabs Bar is used with **anchor links** the semantics and keyboard behaviour should be adapted. 

Out of scope
---
- Tabs component is not affected
- No separate component, stick with tabsbar

Note
---
- links should be navigable by tab key only (not with arrow keys)
- the semantics should follow standard navigation guidelines, see WCAG best practices
- add validation for supported nesting and correct usage of ONLY <a> OR <button> (no mixing)

Also the documentation should clarify exactly those different usage scenarios.

Out of scope
---
- No handling of mixed `a` and `button` elements

Acceptance Criteria
---
- [x] TabsBar with nested links is implemented accessible compliant and with expected interaction
- [x] Validation is implemented

o-hook:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2040/

#### Scope

#### Resolved Issue

Resolves #2040


packages/components/src/components/tabs-bar/tabs-bar.tsx
```diff
@@ -153,7 +156,7 @@ export class TabsBar {
     return (
       <PrefixedTagNames.pScroller
         class="scroller"
-        aria={{ role: 'tablist' }}
+        aria={{ role: this.tabsAreButtons ? 'tablist' : null }}
```
denyo, line 159: This wouldn't set the aria prop at all and therefore not do unneeded loops, validations, etc. in `p-scroller`.
```suggestion
        {...(this.tabsAreButtons && { aria: { role: 'tablist' } })}
```

packages/components/src/utils/validation/throwIfChildrenAreNotEqualOfKind.ts
```diff
@@ -0,0 +1,22 @@
+import { getTagNameWithoutPrefix, throwException } from '..';
+
+export const throwIfChildrenAreNotEqualOfKind = (host: HTMLElement, allowedTagNames: string[]): void => {
```
denyo, line 3: I see we already have `throwIfChildrenAreNotOfKind()` which uses a utility function called `areAllChildrenOfKind()` which is used for custom element use cases with optional prefixing.
Now this new utility does essentially the same but without the custom element support, so naming is kind of confusing. 
I guess those functions should probably be renamed, aligned and/or refactored so we don't have 2 functions which are 90% identical but throw different exceptions.

packages/components/src/components/tabs-bar/tabs-bar.tsx
```diff
@@ -190,6 +198,7 @@ export class TabsBar {
 
   private setTabElements = (): void => {
     this.tabElements = getHTMLElements(this.host, 'a,button');
```
denyo, line 200: I think validation should happen here, otherwise later added children are not validated anymore (see `onSlotchange`).
Perhaps it would be simpler to create a new function similar to `getOnlyChildOfKindHTMLElementOrThrow()` which calls `getHTMLElements()`, does the validation and returns the results if validation passes.

packages/components/src/components/tabs-bar/tabs-bar.tsx
```diff
@@ -190,6 +198,7 @@ export class TabsBar {
...
+    this.tabsAreButtons = this.tabElements.length > 0 && this.tabElements[0].tagName.toLowerCase() === 'button';
```
denyo, line 201:
```suggestion
    this.tabsAreButtons = this.tabElements[0]?.tagName === 'BUTTON';
```

packages/components/src/components/tabs-bar/tabs-bar.tsx
```diff
@@ -89,6 +90,7 @@ export class TabsBar {
   private scrollerElement: HTMLPScrollerElement;
   private direction: ScrollerDirection = 'next';
   private hasPTabsParent: boolean;
+  private tabsAreButtons: boolean;
```
denyo, line 93: Perhaps rename since booleans usually start with the verb.
```suggestion
  private areTabsButtons: boolean;
```

packages/components-react/projects/react-ssr-wrapper/tests/unit/specs/__snapshots__/react-ssr-wrapper.spec.tsx.snap
```diff
@@ -1158,7 +1158,6 @@ exports[`manual test cases should pass internal props correctly to p-tabs-item 1
               >
                 <div
                   class="scroll-wrapper"
-                  role="tablist"
```
denyo, line 1161: Shouldn't this attribute stay since the children are `button` elements? I believe also the other scenario with `a` elements is missing?

packages/components/src/utils/validation/throwIfChildrenAreNotEqualOfKind.ts
o-hook: Hm, I don't see here too much redundant behavior. 
- `throwIfChildrenAreNotOfKind()` does test against the allowance of dedicated custom child elements.
- `throwIfChildrenAreNotEqualOfKind()`does test against the sameness of children, not the allowance (though the exception text takes care about allowed tags)

So maybe it would be more clear if we rename from `throwIfChildrenAreNotEqualOfKind()` to `throwIfChildrenAreNotEqual()`?

packages/components/src/components/tabs-bar/tabs-bar.tsx
o-hook: Yes, I also thought about the missing validation on `slotChange`. Then I decided to only validate on `disconnectedCallback()` otherwise the error was always exposed twice in the log because `slotChange` does also fire on first load?

o-hook: refactored error handling

packages/components/src/utils/validation/throwIfChildrenAreNotEqualOfKind.ts
o-hook: refactored error handling

packages/components-react/projects/react-ssr-wrapper/tests/unit/specs/__snapshots__/react-ssr-wrapper.spec.tsx.snap
o-hook: Manual test for `Tabs Bar` with slotted `a` tags created. 
`role="tablist"` won't be fixed because it has no visual effect.

denyo (approved):

PR #2724, opened 2023-08-09
feat(table): add option to render sticky column headers

### LINKED ISSUE #1947: Sticky Table Header
Scope
---
In contexts of long tables, e.g. My Porsche Tripdata or Sales Cockpit the usability of the table component can be increased if the header is fixed once the contend is scrolled out of view.

Notes  
---
- https://porschedev.slack.com/archives/CT7AVHTTQ/p1656511661874899
- https://css-tricks.com/position-sticky-and-table-headers/
- Example: https://nordhealth.design/components/table/

Acceptance criteria
---
- [ ] Sticky header for table component is enabled

Subtasks
---
- [ ] Task

Rouby:
### Pull Request Checklist

#### Scope

This pull requests adds rudimentary / POC ability to p-table to enable sticky column headers.

#### Resolved Issue

Resolves #1947

### DIFF
--- a/packages/components/src/components.d.ts
+++ b/packages/components/src/components.d.ts
@@ -1193,6 +1193,10 @@ export namespace Components {
          */
         "gradientColorScheme"?: ScrollerGradientColorScheme;
         /**
+          * Limits the height of the scrolling area.
+         */
+        "maxHeight"?: string;
+        /**
           * @deprecated since v3.0.0, will be removed with next major release, use `alignScrollIndicator` instead. Sets the vertical position of scroll indicator
          */
         "scrollIndicatorPosition"?: ScrollerScrollIndicatorPosition;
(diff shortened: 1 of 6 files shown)


stephanschroeter:
With this PR the Table component would support a sticky header but loose it's functionality to be horizontally scrollable in case not enough viewport width is given, which is the case most likely on mobile views. How shall we deal with it?

Rouby:
> With this PR the Table component would support a sticky header but loose it's functionality to be horizontally scrollable in case not enough viewport width is given, which is the case most likely on mobile views. How shall we deal with it?

Oh, I might have missed that. What change is responsible for loosing the horizontal scrolling behaviour? I thought I just added the vertical scrolling (if `maxHeight` is provided); in my tests I was still able to scroll horizontal if not enough width was available for the table 🤔

stephanschroeter:
Unfortunately, with the current DOM structure and A11y compliance ruleset, it's not possible to realise.

PR #2765, opened 2023-09-11
Housekeeping/enable auto dark mode storefront

stephanschroeter:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/housekeeping/enable-auto-dark-mode-storefront

#### Scope

#### Resolved Issue

Resolves [IssueNumber]


packages/storefront/src/store.ts
```diff
@@ -19,7 +20,8 @@ const initialState: State = {
   isMenuActive: false,
   isSearchActive: false,
   selectedFramework: 'vanilla-js',
-  theme: 'light',
+  playgroundTheme: (localStorage.getItem('playgroundTheme') as Exclude<Theme, 'auto'>) || 'light',
```
denyo, line 23:
```suggestion
  playgroundTheme: (localStorage.getItem('playgroundTheme') as State['playgroundTheme'],
```

packages/storefront/src/store.ts
```diff
@@ -43,8 +45,12 @@ export default new Vuex.Store({
     setSelectedFramework(state: State, payload: Framework): void {
       state.selectedFramework = payload;
     },
-    setTheme(state: State, payload: Theme): void {
-      state.theme = payload;
+    setPlaygroundTheme(state: State, payload: Exclude<Theme, 'auto'>): void {
+      localStorage.setItem('playgroundTheme', payload);
```
denyo, line 49: Why only persist one of the themes and not both?
Perhaps "syncing" it to state could also be skipped when getters just return it from localStorage.

packages/storefront/src/views/Page.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <p-tabs-bar v-if="hasTabs" :active-tab-index="activeTabIndex" size="medium">
+    <p-tabs-bar :theme="$store.getters.platformTheme" v-if="hasTabs" :active-tab-index="activeTabIndex" size="medium">
```
denyo, line 3: Usually structural directives are listed first for better readability and consistency. 
```suggestion
    <p-tabs-bar v-if="hasTabs" :theme="$store.getters.platformTheme" :active-tab-index="activeTabIndex" size="medium">
```

packages/storefront/src/main.ts
```diff
@@ -40,10 +40,11 @@ Vue.config.ignoredElements = [/p-\w*/];
 Vue.use({
   install(vue: VueConstructor) {
     vue.component('A11yIcon', A11yIcon);
-    vue.component('ColorBadge', ColorBadge);
     vue.component('OpenBugTemplateInStackBlitz', OpenBugTemplateInStackBlitz);
     vue.component('PartialDocs', PartialDocs);
     // eslint-disable-next-line vue/multi-word-component-names
+    vue.component('Notification', Notification);
```
denyo, line 46: Eslint annotation in previous line shouldn't be needed since `Notification` is a single word.

packages/storefront/src/components/Menu.vue
```diff
@@ -1,24 +1,21 @@
 <template>
   <p-button-pure
+    :theme="$store.getters.platformTheme"
```
denyo, line 3: Structural directive should go first.

packages/storefront/src/components/CyclePlatformTheme.vue
```diff
@@ -0,0 +1,58 @@
...
+    public icon(): 'moon' | 'sun' {
+      if ((this.theme() === 'auto' && this.isPreferredColorSchemeDark()) || this.theme() === 'dark') {
+        return 'sun';
+      } else {
+        return 'moon';
+      }
+    }
+
+    public cycleTheme(): void {
+      const themes: Theme[] = ['auto', ...[this.isPreferredColorSchemeDark() ? 'light' : 'dark']];
```
denyo, line 45:
```suggestion
      const themes: Theme[] = ['auto', this.isPreferredColorSchemeDark() ? 'light' : 'dark'];
```

packages/storefront/src/components/CyclePlatformTheme.vue
```diff
@@ -0,0 +1,58 @@
+<template>
+  <p-button-pure :theme="theme()" :icon="icon()" hide-label="true" @click="cycleTheme()"
```
denyo, line 2: Could use getters instead of functions.

packages/storefront/src/components/ComponentMetaOverview.vue
```diff
@@ -2,10 +2,12 @@
   <div>
     <div style="display: flex; justify-content: space-between; margin: 0 0 1rem">
       <div>🚫 = deprecated<br />🛠 = breakpointCustomizable</div>
-      <p-switch @update="toggleProps" :checked="isToggled">Show all prop values</p-switch>
+      <p-switch :theme="$store.getters.platformTheme" @update="toggleProps" :checked="isToggled"
```
denyo, line 5: When is theme retrieved inline from $store like this and when is there are dedicated getter function? Currently it seems to be random mix.

packages/storefront/src/main.ts
stephanschroeter: True, but for unknown reasons it's complaining:
Screenshot 2023-09-13 at 09 43 47

packages/storefront/src/store.ts
stephanschroeter: Because the `platformTheme` works differently than `playgroundTheme`. `platformTheme` by default uses the auto dark mode and depending on the OS settings of the user the playgroundTheme may be shown in dark or light. Depending on that individual case the option to cycle to a static theme light or dark is chosen.

Imagine I've set my OS settings to theme=auto and I visit the page at night. Then I'll see the page with dark theme. Then I click on cycle theme. Then I would see the page in light theme. If I then would store this setting in local storage (theme=light), exit the page and visit it again on the next day. Then I would see a light page, and if I would then click on cycle theme, I would use theme=auto as next step which again would produce a light page. I hope it's somehow understandable. :)

packages/storefront/src/components/ComponentMetaOverview.vue
stephanschroeter: I suggest, using it inline when it's only used once per file. As soon as it's used twice then we can abstract it into a getter function. I'll refactor it.

denyo (approved):

PR #2801, opened 2023-10-06
Improve Initial Styles & Auto Theme Docs

stephanschroeter:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/housekeeping/initial-styles/

#### Scope

#### Resolved Issue

Resolves [IssueNumber]


packages/components-js/projects/partials/tests/unit/specs/__snapshots__/getInitialStyles.spec.ts.snap
```diff
@@ -101,149 +101,136 @@ p-wordmark {
...
+    p-table,
+    p-tabs,
+    p-text,
+    p-text-field-wrapper,
+    p-text-list,
+    p-textarea-wrapper
+  )
+  a {
   color: currentcolor !important;
+  transition: background var(--p-transition-duration, 0.24s) ease !important;
```
denyo, line 130: Isn't this only relevant when `:hover` is supported?

stephanschroeter: I think we should have the background transition defined in any case, since it does not harm but otherwise produces more CSS especially in the initial styles partial.

denyo (approved):

PR #2818, opened 2023-10-11
[BUG] - Anchor tags inside opened accordion if used in a flyout are focusable/tabbable though it's hidden (Issue/2799)

### LINKED ISSUE #2799: [BUG] - Anchor tags inside opened accordion if used in a flyout are focusable/tabbable though it's hidden
### Bug description
If the `accordion` component with an anchor link is used as a slot in the `flyout` component, this anchor link is still focusable in the following scenario:

- `flyout`is hidden
- `accordion` has state `open`

This only occurs on links slotted inside an `accordion`. If just an anchor tag is provided as a slot inside the `flyout` it doesn't get focused. Also the `accordion` itself doesn't get focused, though it's a focusable element.

### JS Framework

all

3.7.0

all

### Browser version

latest

### Notes
- `inert` attribute (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert) might be a solution to solve the issue 

### Reproduction Example

https://stackblitz.com/edit/twyku1?file=index.html

1. Go to Stackblitz example
2. Tab the links and look at the missing focus and the link url preview in the bottom of your browser
3. You will see, that the anchor link inside the flyout > accordion gets focused even when the flyout is hidden

### Expected behavior

All interactive element inside a flyout are not focusable if the flyout is hidden / `<p-flyout open="false" …`.

denyo:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2799/

#### Scope

#### Resolved Issue

Resolves #2799

### DIFF
--- a/packages/components-js/tests/e2e/puppeteer/specs/__snapshots__/default-dom.e2e.ts.snap
+++ b/packages/components-js/tests/e2e/puppeteer/specs/__snapshots__/default-dom.e2e.ts.snap
@@ -137,7 +137,7 @@ exports[`should have no basic DOM regression for p-flex-item 1`] = `
 `;
 
 exports[`should have no basic DOM regression for p-flyout 1`] = `
-"<div class="root" role="dialog" aria-modal="true" aria-hidden="true" tabindex="-1">
+"<div class="root" role="dialog" aria-modal="true" aria-hidden="true" tabindex="-1" inert="">
   <div class="header">
     <p-button-pure class="dismiss hydrated">
       Dismiss flyout</p-button-pure>
@@ -270,7 +270,7 @@ exports[`should have no basic DOM regression for p-marque 1`] = `
 `;
 
 exports[`should have no basic DOM regression for p-modal 1`] = `
-"<div class="root" role="dialog" aria-modal="true" aria-hidden="true" tabindex="-1">
+"<div class="root" role="dialog" aria-modal="true" aria-hidden="true" tabindex="-1" inert="">
   <p-button-pure class="dismiss hydrated">
     Dismiss modal</p-button-pure>
   <div class="content">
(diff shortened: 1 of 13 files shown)


packages/components/src/components/modal/modal.tsx
```diff
@@ -158,6 +158,9 @@ export class Modal {
             ...parseAndGetAriaAttributes(this.aria),
           })}
           tabIndex={-1}
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          /* @ts-ignore */
+          inert={this.open ? null : true} // prevents focusable elements within nested open accordion
```
stephanschroeter, line 163: Wouldn't it be better to set the attribute on the host element to be absolutely sure everything inside the component can't be focused?

packages/components-js/tests/e2e/puppeteer/specs/__snapshots__/default-dom.e2e.ts.snap
```diff
@@ -137,7 +137,7 @@ exports[`should have no basic DOM regression for p-flex-item 1`] = `
 `;
 
 exports[`should have no basic DOM regression for p-flyout 1`] = `
-"<div class="root" role="dialog" aria-modal="true" aria-hidden="true" tabindex="-1">
+"<div class="root" role="dialog" aria-modal="true" aria-hidden="true" tabindex="-1" inert="">
```
o-hook, line 140: Could you get rid of the empty value? `inert` has no value at all and it looks a bit weird with empty "".

denyo: I don't think so. That's in the nature of jsx.

packages/components/src/components/modal/modal.tsx
denyo: Probably makes no difference but adds the chance that the attribute can be removed.

stephanschroeter (approved):

Issue #2642, opened 2023-07-05
[Feature Request] - Button should accept `name` and `value`


Rouby:
Scope
---
Button and Button Pure shall support `name` and `value` as attribute/prop since it's HTML default behaviour.

Notes  
---
- The default HTML element `button` supports setting a `name` and `value`. This can be used e.g. in forms to determine which submit button was pressed, see: https://www.w3schools.com/tags/att_button_value.asp
- Button, Button Pure and (Button Tile?) has to forward those attributes/props to the underlying button element and use them in case a parent form is getting submitted by the button.
- Evaluate if `type="reset"` works for Button and Button Pure and if it's functionality is covered by test automation

Acceptance criteria
---
- [ ] Button and Button Pure supports `name` and `value` attributes/props

PR #2862, opened 2023-10-30
Button should accept name and value (issue/2642)

henrijoss:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2642

#### Scope

#### Resolved Issue

Resolves [#2642](https://github.com/porsche-design-system/porsche-design-system/issues/2642)


packages/components/src/components/button-pure/button-pure.tsx
```diff
@@ -115,7 +123,9 @@ export class ButtonPure {
     improveButtonHandlingForCustomElement(
       this.host,
       () => this.type,
-      () => this.isDisabledOrLoading
+      () => this.isDisabledOrLoading,
+      () => this.name,
```
denyo, line 127: Should these 2 now be rendered on the inner `button` for transparency and maybe a11y, too?

packages/components/src/utils/button-handling.spec.ts
```diff
@@ -0,0 +1,80 @@
+import * as handleButtonUtils from './button-handling';
+import { handleButtonEvent, improveButtonHandlingForCustomElement } from './button-handling';
+
+describe('improveButtonHandlingForCustomElement()', () => {
+  test('it should add a click event listener to the element', () => {
```
denyo, line 5: I think we always use the `it` alias instead of `test` and then the description starts with `should..`.

packages/components/src/utils/button-handling.ts
```diff
@@ -1,19 +1,25 @@
...
-import { getClosestHTMLElement } from './dom';
+import { getClosestHTMLElement, setAttributes } from './dom';
 
 export const improveButtonHandlingForCustomElement = (
   element: HTMLElement,
   getType: () => ButtonType,
-  getDisabled: () => boolean
+  getDisabled: () => boolean,
+  getName?: () => string | undefined,
+  getValue?: () => string | undefined
```
denyo, line 9: Could probably just capture these 4 with the rest operator like `...args` and then spread them on `handleButtonEvent` since the order is identical.

packages/components/src/utils/button-handling.ts
```diff
@@ -1,19 +1,25 @@
...
+  );
 };
 
 export const handleButtonEvent = (
   event: MouseEvent | KeyboardEvent,
   element: HTMLElement,
   getType: () => ButtonType,
-  getDisabled: () => boolean
+  getDisabled: () => boolean,
+  getName?: () => string | undefined,
```
denyo, line 21: Why optional? Would be better typed and safer if it was always necessary.
Then we wouldn't need optional chaining down below, too.

henrijoss: Since it is used in the `multi-select` and `text-field-wrapper` without those params

packages/components/src/utils/button-handling.ts
henrijoss: Doesn't really make it easier because of typing

denyo (approved):

Issue #2820, opened 2023-10-12
Add validation for different versions of PDS


denyo:
Scope
---
In order to increase awareness and ultimately improve loading behavior and performance across all Porsche web apps/sites we should extend our validation that occurs during initialization.  

Currently this validation covers the following topics
- `validateGetInitialStylesUsage`
- `validateGetFontFaceStylesheetUsage`
- `validateGetFontLinksUsage`

The idea is to add another validation that checks the different versions of PDS which should ideally match for all prefixes.  
If this is not the case, a meaningful warning should be output.

Out of Scope
---
List topics that are not part of this task.

Notes  
---
- Evaluate how PDS updates are integrated regularly within applications and widgets at the same time (and publish it on the Storefront).

Acceptance criteria
---
- [ ] A clear validation message is shown, if multiple versions (>1) of PDS components are used in a micro-frontend application 

Subtasks
---
- [ ] Task

PR #2867, opened 2023-11-02
Add validation for different versions of PDS (Issue/2820)

henrijoss:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2820

#### Scope

#### Resolved Issue

Resolves [#2820](https://github.com/porsche-design-system/porsche-design-system/issues/2820)
Item statusOpen


packages/components/src/types/common-types.d.ts
```diff
@@ -27,11 +27,27 @@ export type { LinkTarget } from '../utils/link-button/link-target';
...
+    prefixes: string[]; // to not break older versions
+  };
+};
+
 // ROLLUP_REPLACE_VARIABLES are replaced via rollup
 declare global {
   const ROLLUP_REPLACE_IS_STAGING: string;
   const ROLLUP_REPLACE_VERSION: string;
   const ROLLUP_REPLACE_CDN_BASE_URL: string;
+  // eslint-disable-next-line @typescript-eslint/consistent-type-definitions
```
denyo, line 47: Do we still need this annotation after the types have been aligned?

packages/components/src/utils/validation/validateVersions.ts
```diff
@@ -0,0 +1,49 @@
...
+ * Validates the versions of the Porsche Design System and logs a warning if multiple different versions are detected.
+ * If multiple versions are found, it recommends upgrading to the latest used version.
+ */
+export const validateVersions = (): void => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const { cdn, ...versions } = document.porscheDesignSystem;
+  if (Object.keys(versions).length > 1) {
+    const sortedVersions = Object.entries(versions)
+      .map(([version, { prefixes }]) => [version, prefixes])
+      .sort((a: [Version, string[]], b: [Version, string[]]) => sortVersions(a[0], b[0]));
```
denyo, line 16: Could probably move parameter typings into the `.map<T>` generic.

packages/components/src/utils/validation/validateVersions.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+});
+
+describe('sortVersions()', () => {
+  it.each([
+    ['2.0.0', '1.0.0', 1],
+    ['1.1.0', '1.0.0', 1],
+    ['1.0.1', '1.0.0', 1],
+    ['2.1.2', '2.1.1', 1],
+    ['2.1.2-rc.2', '2.1.2-rc.1', 1],
+  ])('(%s) > (%s) = %d', (versionA: Version, versionB: Version, expected: number) => {
```
denyo, line 91: Parameter types could be inherited from `it.each<T>` generic. This would probably add type safety to the arrays, too.

packages/components/src/utils/validation/validateVersions.ts
```diff
@@ -0,0 +1,49 @@
...
+  const a = versionA.replace('-rc', '').split('.').map(Number);
+  const b = versionB.replace('-rc', '').split('.').map(Number);
+
+  for (let i = 0; i < a.length; i++) {
+    if (a[i] < b[i]) {
+      return -1;
+    }
+    if (a[i] > b[i]) {
+      return 1;
+    }
```
denyo, line 45: Why not use an else or if else case and handle all returns here? Perhaps even with ternary operator.

denyo:
Do we need an e2e test covering this and seeing it in action? Could be done similar to the crawler e2e test where another version is initialized via hardcoded partial.

packages/components/src/types/common-types.d.ts
henrijoss: Yes since its an interface, maybe we could change it to a type?

packages/components/src/utils/validation/validateVersions.ts
henrijoss: Because only if none of the comparisons return the versions are equal and it returns 0.

packages/components-js/tests/e2e/puppeteer/specs/validate-versions.e2e.ts
```diff
@@ -0,0 +1,66 @@
...
+
+let page: Page;
+beforeEach(async () => {
+  page = await browser.newPage();
+  initConsoleObserver(page);
+});
+afterEach(async () => await page.close());
+
+it('should show warning about multiple different versions correctly', async () => {
+  enableBrowserLogging(page);
```
denyo, line 20: Can be removed since it is just for development?

packages/components-js/tests/e2e/puppeteer/specs/validate-versions.e2e.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  expect(porscheDesignSystem[version]).toBeDefined();
+  expect(porscheDesignSystem['3.7.0']).toBeDefined();
+  expect(porscheDesignSystem['3.7.0'].prefixes).toEqual(prefixes);
+
+  const versionWarning = getConsoleWarnings().find((warning) => warning.text().includes('Multiple different versions'));
+
+  expect(versionWarning).toBeDefined();
+
+  const warningArgs = await Promise.all(versionWarning.args().map(async (arg) => await arg.evaluate((arg) => arg)));
```
denyo, line 43: Let's use `getConsoleWarnings()` for consistency.

packages/components-js/tests/e2e/puppeteer/specs/validate-versions.e2e.ts
```diff
@@ -0,0 +1,66 @@
...
+  enableBrowserLogging(page);
+
+  const prefixes = ['test', 'my-prefix'];
+
+  await setContentWithDesignSystem(
+    page,
+    `<p-text>Some Text</p-text>
+    ${getOldLoaderScriptForPrefixes(prefixes)}`,
+    {
+      withoutWaitForComponentsReady: true,
```
denyo, line 29: Can we get rid of this to actually verify that things still work?

packages/components/src/setup.ts
```diff
@@ -9,4 +9,5 @@ export default (): void => {
 
   injectGlobalStyle();
   validatePartialUsage();
+  validateVersions();
```
denyo, line 12: I am wondering this is the best place, perhaps `loadComponentLibrary()` would work better since prefix collisions are checked there.
Also, this will only reliably work when it is executed last, right? So if this version is initialized first, it will miss later ones.
How does this behave in a real application, e.g. porsche.com or even our sample integrations?

packages/components-js/tests/e2e/puppeteer/specs/validate-versions.e2e.ts
henrijoss: I already use getConsoleWarnings but i need the args separately because if i just convert the ConsoleMessage into text i can't check the object argument properly.

denyo (dismissed):

packages/components/src/types/common-types.d.ts
```diff
@@ -27,11 +27,27 @@ export type { LinkTarget } from '../utils/link-button/link-target';
 
 export type { AlignLabel, AlignLabelDeprecated } from '../utils/link-button/align-label';
 
+export type PorscheDesignSystem = {
```
stephanschroeter, line 30: Isn't this type defined twice within the mono repo now?

packages/components/src/utils/validation/validateVersions.ts
```diff
@@ -0,0 +1,29 @@
...
+ * More information about other implementations can be found in the PR: https://github.com/porsche-design-system/porsche-design-system/pull/2867
+ */
+export const validateVersions = (): void => {
+  // Uses a timeout to increase the chances that all used versions are loaded and initialized
+  setTimeout(() => {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { cdn, ...versions } = document.porscheDesignSystem;
+    if (Object.keys(versions).length > 1) {
+      consoleWarn(
+        'Multiple different versions detected!\nPlease upgrade all instances to the latest version in use.\nRefer to the document.porscheDesignSystem object for detailed information on the current versions in use.\n',
```
stephanschroeter, line 24: Maybe we should add a small info why different versions are bad. "Please upgrade all instances to the latest version in use for the best performance."

packages/components/src/utils/log/logger.ts
```diff
@@ -3,7 +3,7 @@ const prefix =
     ? `[Porsche Design System v${ROLLUP_REPLACE_VERSION}]` // this part isn't covered by unit tests
     : '[Porsche Design System]';
 
-export const consoleWarn = (...messages: string[]): void => {
+export const consoleWarn = (...messages: (string | object)[]): void => {
```
stephanschroeter, line 6: Is the object type needed?

packages/components/src/utils/validation/validateVersions.ts
stephanschroeter, line 24:
```suggestion
        'Multiple different versions detected!\nWhile bootstrapping multiple versions is valid, it's highly recommended to upgrade all instances to the latest version in use for the best performance.\nRefer to the document.porscheDesignSystem object for detailed information on the current versions in use.\n',
```

packages/components/src/utils/log/logger.ts
henrijoss: Yes in order to pass in the document.porscheDesignSystem object

stephanschroeter (approved):

PR #2885, opened 2023-11-13
Select options get displayed in the background while being used within a table (Issue/2725)

### LINKED ISSUE #2725: Select options get displayed in the background while being used within a table
### Contact Details

extern.aleksander.bujko@porsche.de

### JS Framework

Angular

3.5.0.

Mobile Firefox, Chrome

### Browser version

_No response_

### Reproduction Example

https://stackblitz.com/edit/angular-xnmnsp-fszw7y?file=src%2Fapp%2Fapp.component.html

1. Create a table using the Design System Component
2. Put a select field with some options within a table cell
3. Navigate to the latest table row 
4. If you open the select field the options cannot be selected, since they are getting displayed in the background
image

### Expected behavior

It is possible to use a select field within a table, if it's located in the last row. 

### Notes

- https://github.com/tailwindlabs/tailwindcss/discussions/2672
- https://play.tailwindcss.com/jPbHqRyZN5

o-hook:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2725/

#### Scope

#### Resolved Issue

Resolves #2725


packages/components/src/components/select-wrapper/select-wrapper-dropdown/select-wrapper-dropdown-utils.spec.ts
```diff
@@ -342,3 +343,69 @@ describe('getDropdownVisibility()', () => {
...
+          left: 10,
+          top: 20,
+          bottom: 50,
+          right: 110,
+        }) as DOMRect
+    );
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
```
denyo, line 385: I think that's covered by `jest.config.js`.

packages/components/src/components/select-wrapper/select-wrapper-dropdown/select-wrapper-dropdown-utils.ts
```diff
@@ -1,5 +1,6 @@
 import type { DropdownDirectionInternal } from '../select-wrapper/select-wrapper-utils';
-import { getHTMLElements, getTagName, hasAttribute } from '../../../utils';
+import { determineDropdownDirection, getHTMLElements, getTagName, hasAttribute } from '../../../utils';
+import { SelectWrapperDropdownDirection } from '../select-wrapper/select-wrapper-utils';
```
denyo, line 3:
```suggestion
import type { SelectWrapperDropdownDirection } from '../select-wrapper/select-wrapper-utils';
```

packages/components/src/components/select-wrapper/select-wrapper-dropdown/select-wrapper-dropdown.tsx
```diff
@@ -182,56 +207,68 @@ export class SelectWrapperDropdown {
...
-                    </li>,
-                  ];
-                })
-              )}
-            </ul>
-          ),
+          <div
+            {...(this.isNativePopover && {
+              popover: 'auto',
+              id: 'popover',
```
denyo, line 213: Let's share this id via const.

packages/components/src/components/select-wrapper/select-wrapper-dropdown/select-wrapper-dropdown.tsx
```diff
@@ -182,56 +207,68 @@ export class SelectWrapperDropdown {
...
+              class: 'popover',
+              ...(this.popoverElement?.matches(':popover-open') && {
+                'popover-open': true,
+              }),
+            })}
+            ref={(el) => (this.popoverElement = el)}
+          >
+            {this.isOpen && (
+              <ul
```
denyo, line 222: Can't this be made a native popover when needed? 
Or do we always need the wrapping div?

packages/components/src/components/select-wrapper/select-wrapper-dropdown/select-wrapper-dropdown.tsx
```diff
@@ -86,11 +94,26 @@ export class SelectWrapperDropdown {
...
+          this.host,
+          this.optionMaps,
+          this.popoverElement,
+          this.buttonElement,
+          this.direction
+        );
+        addNativePopoverScrollAndResizeListeners(this.host, this.parentTableElement, this.popoverElement);
+      }
+    } else if (this.isNativePopover) {
+      this.popoverElement.hidePopover();
```
denyo, line 116: Same could probably be achieved like this:
```suggestion
    } else {
      this.popoverElement?.hidePopover();
```

packages/components/src/components/select-wrapper/select-wrapper-dropdown/select-wrapper-dropdown-utils.spec.ts
o-hook: Fixed

packages/components/src/components/select-wrapper/select-wrapper-dropdown/select-wrapper-dropdown-utils.ts
o-hook: Fixed

packages/components/src/components/select-wrapper/select-wrapper-dropdown/select-wrapper-dropdown.tsx
o-hook: I tried this first, but unfortunately the `ul` element doesn't exist until the list is in open state. And the browser can't do its portal "magic" anymore. So we need the wrapping div with `popover` attribute.

packages/components/src/components/select-wrapper/select-wrapper-dropdown/select-wrapper-dropdown.tsx
o-hook: Fixed

packages/components/src/components/select-wrapper/select-wrapper-dropdown/select-wrapper-dropdown-styles.spec.ts
```diff
@@ -59,39 +59,46 @@ describe('getFilterStyles()', () => {
...
+    ['down', 'light', true],
+    ['down', 'dark', false],
+    ['down', 'dark', false],
+    ['up', 'light', false],
+    ['up', 'light', true],
+    ['up', 'light', false],
+    ['up', 'dark', false],
+    ['up', 'dark', false],
+  ])(
+    'should return correct css for direction: %s, isOpen: %s and theme: %s, isNativePopover: %s',
```
denyo, line 73: 3 parameters are defined in array but 4 are mentioned here so besides mixup the last one will always be undefined.

packages/components/src/components/select-wrapper/select-wrapper-dropdown/select-wrapper-dropdown.tsx
```diff
@@ -119,7 +144,8 @@ export class SelectWrapperDropdown {
       this.state,
       this.disabled,
       this.filter,
-      this.theme
+      this.theme,
+      this.isNativePopover
```
denyo, line 148: Let's keep `theme` as last parameter like everywhere else.

packages/components/src/components/select-wrapper/select-wrapper-dropdown/select-wrapper-dropdown-utils.ts
```diff
@@ -161,3 +161,19 @@ export const getDropdownVisibility = (
...
+  nativePopover: HTMLElement,
+  direction: SelectWrapperDropdownDirection
+): void => {
+  const { left, top, width, height } = host.getBoundingClientRect();
+  const isDirectionDown =
+    direction === 'down' ||
+    (direction === 'auto' &&
+      determineDropdownDirection(host, getAmountOfVisibleOptionsAndOptgroups(optionMaps)) === 'down');
+  nativePopover.style.left = `${left + window.scrollX}px`;
+  nativePopover.style.top = `${top + window.scrollY + (isDirectionDown && height)}px`;
```
henrijoss, line 177:
```suggestion
  nativePopover.style.top = `${top + window.scrollY + (isDirectionDown ? height : 0)}px`;
```
Maybe this would be better

henrijoss:
Storefront example missing? At least we have one for the popover

o-hook:
> Storefront example missing? At least we have one for the popover

I don't think so that this is needed. On the Storefront there should be only examples which show how to use and implement components and not how they work (out of the box) in specific contexts or environments.

denyo (approved):

Issue #2863, opened 2023-10-31
[BUG] - Flyout has two scrollbars & Modal has hidden scrollbar on specific OS/Browser combination/settings


o-hook:
### Bug Description
On Windows machines with MS Edge or Chrome Browser or on macOS with always visible scrollbars, the `Flyout` component is rendered with two scrollbars on the right/left side if its contents are scrollable. At the same time a `Modal`'s scrollbar might not even be visible at all.

![flyout](https://github.com/porsche-design-system/porsche-design-system/assets/40788023/60ea1227-f4d2-4b61-a643-80e027a9f0b5)

![Image](https://github.com/porsche-design-system/porsche-design-system/assets/30120450/45ee3074-1eb1-4a3c-9da2-a24f62ef0fed)

### Contact Details

Slack: https://porschedev.slack.com/archives/CT7AVHTTQ/p1698674271270639

### JS Framework

All

### Porsche Design System version

3.8.0

### What browsers are you seeing the problem on?

Chrome, Microsoft Edge

### Browser version

latest

### Notes

- Flyout issue is caused because of `overflow-y: scroll` is set on `body`.
- Modal issue is caused because scrollbar is behind backdrop
- Try to enforce scrollbars in Playwright VRT test to cover correct scroll behaviour
- Maybe the "advanced scroll lock" should only be applied for iOS in addition, because with the current solution the page might jump horizontally too depending on OS scroll settings when opening a Modal. On iOS we can be sure, that the scrollbar is always overlayed.
- Related #2869 

### Steps to reproduce the behavior

**In Windows (11) and Microsoft Edge/Chrome:**
**On macOS: Set `Show scroll bars` to `always` in OS settings:**

1. Go to https://designsystem.porsche.com/v3/components/flyout/examples
2. Open Flyout
3. See error (two scrollbars)

---

1. Go to https://designsystem.porsche.com/v3/components/modal/examples
2. Open Modal
3. See error (no scrollbar visible)

### Expected behavior

- Only one scrollbar is shown if the Flyout is scrollable
- Scrollbar of Modal is visible when content is scrollable
- Scrollbar is scrollable by mouse pointer only in addition to scroll wheel and touch move for Flyout and Modal

stephanschroeter:
![Image](https://github.com/porsche-design-system/porsche-design-system/assets/30120450/123e15b7-74f0-472e-a995-aedc9f767115)
[...]

PR #2907, opened 2023-11-27
[BUG] - Flyout has two scrollbars & Modal has hidden scrollbar on specific OS/Browser combination/settings (issue/2863)

henrijoss:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2863

#### Scope

#### Resolved Issue

Resolves [#2863](https://github.com/porsche-design-system/porsche-design-system/issues/2863)


packages/components/src/components/modal/modal-styles.ts
```diff
@@ -238,6 +243,7 @@ export const getComponentCss = (
         left: '8px',
         display: 'flex',
         justifyContent: 'flex-end',
+        zIndex: 1, // To be on top of transformed full screen content
```
denyo, line 246: Comment is somewhat wrong or at least confusing.
I guess it is related to full width and `stretchToFullModalWidthClassName`?

packages/components/src/utils/focusTrap.ts
```diff
@@ -35,6 +35,7 @@ export let documentKeydownListener: (e: KeyboardEvent) => void;
 export const setFocusTrap = (
   host: HTMLElement,
   isOpen: boolean,
+  firstElementChild: HTMLElement,
```
denyo, line 38: Can we use a better name if this is the only way to fix the issue?

denyo (dismissed):

denyo (approved):

Issue #2836, opened 2023-10-19
Opt-out for Text Field Wrapper's submit of `input type="search"`


denyo:
Scope
---
Currently the `p-text-field-wrapper` with `input type="search"` has a conditional behavior depending on wether or not the component is within a `form` element.

Outside of a form, there is an icon on the left side and no submit button/functionality.  
Within a form, there is a submit button on the right side.

![Screenshot 2023-10-19 at 13 42 34](https://github.com/porsche-design-system/porsche-design-system/assets/4380295/38abb2fe-a6de-430d-b272-696f814ac3d9)

This works for typical search bar scenarios, but does not if there are other `input`s in the same form or you simply don't to have the submit functionality because of an actual `button type="submit"`. 

Therefore, an opt-out would be helpful.

![screenshot_2023-10-19_at_09 33 24_720](https://github.com/porsche-design-system/porsche-design-system/assets/4380295/886912cc-4453-4a8e-a2e7-856ed534d8b9)

Out of Scope
---
List topics that are not part of this task.

Notes  
---
- https://porschedev.slack.com/archives/CT7AVHTTQ/p1697701168738309
- `submitButton: true|false`

Acceptance criteria
---
- [x] Teams should be able to hide the search button individually if the `Textfield Wrapper` is wrapped inside a form.

Subtasks
---
- [ ] Task

PR #2908, opened 2023-11-27
Issue/2836 Opt-out for Text Field Wrapper's submit of input type="search"

tatjana-gartner2:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2836/components/text-field-wrapper/examples#type-search

#### Scope

#### Resolved Issue

Resolves [#2836](https://github.com/porsche-design-system/porsche-design-system/issues/2836)


packages/components/src/components/text-field-wrapper/text-field-wrapper.examples.md
```diff
@@ -250,16 +253,19 @@ export default class Code extends Vue {
     return this.typePassword.replace(/label="Some label"/, '$& show-password-toggle="false"');
   }
 
-  typeSearch =
-`<p-text-field-wrapper label="Some label">
+  submitButton = true;
+  submitButtons = [true, false];
+  get typeSearchMarkup() {
+    return `<p-text-field-wrapper label="Some label" submit-button="${this.submitButton}">
```
o-hook, line 259: This example is a bit confusing: if `submit-button="true"` it does not show a submit button. Though this is intended, it is somehow difficult to understand. Maybe we can enhance the description above to make it more clearer, that the text-field becomes never a submit button if it is used outside a `form` element. Or we delete the prop on the first example?

packages/components/CHANGELOG.md
```diff
@@ -14,6 +14,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### [Unreleased]
 
+#### Added
+
+- Prop `submit-button` for `Text Field Wrapper`
```
denyo, line 19: Would be helpful to add in which context and scenario this is relevant, otherwise I might just add the prop and be surprised that nothing happens.

packages/components-js/src/pages/text-field-wrapper.html
```diff
@@ -195,6 +195,29 @@
         value="some really long password with many words and amazing special characters, letters big and small, numbers"
```
o-hook, line 195: The value text of "Password" is not correct (also on other `type="search"` examples).

denyo (approved):

PR #2914, opened 2023-11-30
Housekeeping/update dependencies

denyo:
### Pull Request Checklist

#### References

#### Scope

-  skipped playwright because it brings`node@20.10` because of playwright docker image which has frequent memory leaks in CI when building
- ~~`ts-node` can't handle latest version of `typescript@5.3.2` since it can't extend `tsconfig.json` from shared package anymore,~~ solution could be to switch from `ts-node` to `tsx`, but `jest` or `ts-jest` is relying on `ts-node` and runs into similar problems
  - https://github.com/TypeStrong/ts-node/issues/2076
  - https://github.com/privatenumber/tsx
- Updated everything except for

Bildschirmfoto 2023-12-18 um 11 47 59


henrijoss (approved):
looks good to me

PR #2956, opened 2024-01-02
Storefront / Improve Props typing (Issue/2821)

### LINKED ISSUE #2821: Storefront / Improve Props typing
Scope
---
Basically for all components that have non primitive props the `Props` section in our documentation is wrong.

Wrong types for e.g. [Link Tile](https://designsystem.porsche.com/v3/components/link-tile/props):
```
type AspectRatio = '1:1' | '4:3' | '3:4' | '16:9' | '9:16'
type Compact = boolean
type Size = 'inherit' | 'default'
type Weight = 'regular' | 'semi-bold' | 'semibold'
type BreakpointCustomizable<T> = { base: T; xs?: T; s?: T; m?: T; l?: T; xl?: T; xxl?: T; }

// within props table
'1:1'
'4:3'
'3:4'
'16:9'
'9:16'
BreakpointCustomizable<AspectRatio>
```

Correct types for e.g. Link Tile (component props first, alphabetically ordered, then common types):
```
type LinkTileAlign = 'top' | 'bottom'
type LinkTileAspectRatio = '1:1' | '4:3' | '3:4' | '16:9' | '9:16'
type LinkTileBackground = 'light' | 'dark' | 'auto'
type LinkTileSize = 'inherit' | 'default'
type LinkTileTarget = '_self' | '_blank' | '_parent' | '_top' | string
type LinkTileWeight = 'regular' | 'semi-bold' | 'semibold'
type BreakpointCustomizable<T> = { base: T; xs?: T; s?: T; m?: T; l?: T; xl?: T; xxl?: T; }

// within props table
LinkTileAspectRatio | BreakpointCustomizable<LinkTileAspectRatio>
```

Notes  
---
- Related: 
  - #1473 
  - #2759
- We should either fix the props transformation script (`cleanReadMes.ts`) using the Stencil generated types or we should generate them completely on our own.
- We should try to add some unit tests
- Maybe we should extend Component Meta with prop type information instead of try to transform auto generated md files

Acceptance criteria
---
- [x] All component prop types in the props section of the Storefront are presented correctly

Subtasks
---
- [x] extend componentMeta by relevant data, e.g. prop descriptions
- [x] render prop tables via componentMeta
- [x] simplify `cleanReadmes.ts` script

denyo:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2821/components/banner/props

#### Scope

#### Resolved Issue

Resolves #2821


packages/storefront/src/components/PropsTable.vue
```diff
@@ -0,0 +1,155 @@
...
+            ...(Array.isArray(meta.allowedValues)
+              ? meta.allowedValues.map(
+                  (val) =>
+                    wrapInCodeCode(
+                      val === 'string'
+                        ? val
+                        : val === null
+                          ? 'undefined'
+                          : typeof val === 'number'
+                            ? `${val}`
```
stephanschroeter, line 107: Do we really want to write code with all of these inline conditions and shorthand syntax, seams hard to debug later on?

packages/storefront/src/components/PropsTable.vue
```diff
@@ -0,0 +1,155 @@
...
+  @use '@porsche-design-system/components-js/styles' as *;
+  @import '../styles/shared.styles';
+
+  :deep(.deprecated) {
+    color: #d5001c;
+    text-transform: uppercase;
+  }
+
+  table,
+  :deep(table) {
```
stephanschroeter, line 142: why do we need `:deep` here?

packages/storefront/src/components/PropsTable.vue
```diff
@@ -0,0 +1,155 @@
+<template>
+  <div>
+    <code v-if="hasBreakpointCustomizableProp" class="code-before-table" v-html="breakpointCustomizableGeneric"></code>
+    <table>
```
stephanschroeter, line 4: we could also use `p-table` now

packages/storefront/src/components/PropsTable.vue
denyo: To apply styling on nested `code` tags that are not rendered via template, but instead `v-html`.
```
@mixin tableStyles {
  margin-top: $pds-spacing-fluid-medium;
  border-collapse: collapse;

  code ~ code::before { // <---
    content: '| ';
  }
  ...
```

packages/storefront/src/components/PropsTable.vue
denyo: Native one gave better result since `p-table` was scrollable with long descriptions.

stephanschroeter: Depending on the circumstances I think it's good when it becomes scrollable, otherwise type definitions might become hyphenated or the table width is larger than the viewport which either zooms the whole page or makes it scrollable horizontally.

packages/storefront/src/components/PropsTable.vue
denyo: Refactored it to `formatPropType()` util with unit tests.

packages/storefront/src/components/PropsTable.vue
denyo: Default `p-table` will look and behave like this which is a step backwards.

Bildschirmfoto 2024-01-08 um 13 26 55

Having it scrollable doesn't really help smaller viewports, too, unless we add support for fixed 1st column.

Bildschirmfoto 2024-01-08 um 13 33 49

I think if we want to use `p-table` here, we should use it everywhere for consistency.

stephanschroeter: Ok, but then we'll definitely need to have a look in a follow up issue because the current table implementation is broken for smaller viewports.

henrijoss (approved):

PR #2957, opened 2024-01-02
Flippable icons in rtl-mode (issue/2804)

### LINKED ISSUE #2804: Improve RTL (right-to-left) support for dedicated icons (mirrored)
Scope
---
Certain icons (e.g. `arrow-left` or `arrow-right`) needs to be mirrored when used in RTL (right-to-left) mode. 

Notes  
---
- https://css-tricks.com/almanac/selectors/d/dir/
- https://caniuse.com/?search=%3Adir
- https://rtlstyling.com/posts/rtl-styling#bidirectional-icons

Acceptance criteria
---
- [x] Relevant icons are mirrored when used in RTL (right-to-left) mode.

Subtasks
---
- [x] Task

stephanschroeter:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2804/components/icon

#### Scope

- Certain icons will flip automatically when used in RTL mode
- Upgraded to latest Playwright version which supports `:dir(rtl)` selector for Chromium
  - makes it necessary to get rid of `concurrently` for assets package (because of new Playwright Docker image which is based on Node >= 20), otherwise build runs into random errors
  - makes it necessary to refactor JSS styles for **Pin Code**, otherwise Shadow Root is not initializing in Webkit browser engine in Playwright environment for VRT

#### Resolved Issue

Resolves #2804


packages/components/src/components/icon/icon-styles.ts
```diff
@@ -141,6 +160,11 @@ export const getComponentCss = (
               height: fontLineHeight,
               font: `${sizeMap[size]} ${fontFamily}`,
             }),
+        ...(isFlippableIcon(name) && {
```
stephanschroeter, line 163: Wouldn't it be easier to maintain to flip all icons and exclude just logos and icons with numbers, e.g. "360"?

packages/components/src/components/pin-code/pin-code-styles.ts
```diff
@@ -45,11 +45,18 @@ export const getComponentCss = (
           opacity: 0.2, // TODO: not in sync with e.g. checkbox/radio-button loading style
           cursor: 'not-allowed',
         }),
-        ...Object.fromEntries(
+        // since @playwright/test@1.40.1 this does not work anymore in Webkit browser engine for unknown reasons
```
denyo, line 48: Should we really change implementations to make tests work?

package.json
```diff
@@ -182,7 +182,7 @@
     "**/workbox-build/rollup": "^4.9.1"
   },
   "volta": {
-    "node": "18.18.1",
+    "node": "20.10.0",
     "yarn": "1.22.19"
```
denyo, line 186: I think this version is also different in the new base image.

o-hook (dismissed):

packages/components/src/components/pin-code/pin-code-styles.ts
stephanschroeter: Well, normally not but I really have no glue why it stopped working. It's also nearly impossible to debug since it only happens with Playwright Webkit browser engine but not with e.g. macOS Safari. 
On the other hand the new implementation might be a bit larger in size but is also more readable and since this component won't be heavily used I think it's okay.

package.json
stephanschroeter: Screenshot 2024-01-03 at 09 52 29
Seems good to me.

o-hook (approved):

Issue #2891, opened 2023-11-15
[BUG] - Pin-code current-input id is not set correctly


henrijoss:
### JS Framework

All

### Porsche Design System version

3.8.0

### Reproduction Example

https://designsystem.porsche.com/v3/components/pin-code/examples

### Steps to reproduce the behavior

1. Go to [https://designsystem.porsche.com/v3/components/pin-code/examples](https://designsystem.porsche.com/v3/components/pin-code/examples)
2. Inspect input elements, current-input id is on first input
3. Type in some value
4. `current-input` id is gone

### Expected behavior

`current-input` id is set on relevant input element:
- no value is entered at all: set `current-input` id on first input element
- some value is entered: set `current-input` id on the first input element which does not have a value
- all inputs have some value: set `current-input` id on the last input element

tatjana-gartner2:
Since whitespaces are added to the value to allow not in sequence typing, the value value must be sanitised in this calculation`id={index === this.value.length ? currentInputId : null}`

PR #2985, opened 2024-01-10
[BUG] - Pin-code current-input id is not set correctly (Issue/2891)

henrijoss:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2891

#### Scope

#### Resolved Issue

Resolves [#2891](https://github.com/porsche-design-system/porsche-design-system/issues/2891)


packages/components/src/components/pin-code/pin-code-utils.spec.ts
```diff
@@ -316,3 +317,40 @@ describe('syncHiddenInput()', () => {
...
+    [2, '123 56', 6, false],
+    [3, '123 56', 6, true],
+    [4, '123 56', 6, false],
+    [5, '123 56', 6, false],
+    [0, '1234', 4, false], // All inputs have some value: set current-input id on the last input element
+    [1, '1234', 4, false],
+    [2, '1234', 4, false],
+    [3, '1234', 4, true],
+  ])(
+    'should return correct current input for index: %d, value: %s and length: %d',
```
tatjana-gartner2, line 350:
```suggestion
    'should for index: %d, value: %s and length: %d' return: %s,
```

packages/components/src/components/pin-code/pin-code-utils.ts
```diff
@@ -110,3 +110,22 @@ export const isFormSubmittable = (host: HTMLElement, form: HTMLFormElement): boo
...
+ * @param {number} index - The index of the current input element.
+ * @param {string} value - The input value.
+ * @param {number} length - The total number of input elements.
+ * @returns {boolean} - True if the current input should have a specific identifier, indicating it corresponds to the first non-filled input; otherwise, false.
+ */
+export const isCurrentInput = (index: number, value: string, length: number): boolean => {
+  if (!value) {
+    return index === 0; // No value entered at all: set current-input id on the first input element
+  }
+  const firstWhitespaceIndex = value.split('').findIndex((char) => char === ' ');
```
denyo, line 126: I think `value.indexOf(' ')` would be easier and shorter.

packages/components/CHANGELOG.md
```diff
@@ -67,6 +67,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 #### Fixed
 
+- `Pin Code`: `current-input` id is not set correctly
```
denyo, line 70: From a consumer perspective I would have no clue what this means. Instead describing the focus behavior that was improved would be better.

denyo:
Should we add e2e tests for the different scenarios that are described in the ticket?

henrijoss:
> Should we add e2e tests for the different scenarios that are described in the ticket?

There already is an e2e test for checking if the label click works when the `current-input` id is set. Testing if the correct input has the id should be covered with the unit test.

denyo (approved):

PR #2989, opened 2024-01-11
Housekeeping/fix storefront navigation

o-hook:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/housekeeping/fix-storefront-navigation

#### Scope

Change active state of sidebar


packages/storefront/src/components/Navigation.vue
```diff
@@ -62,6 +62,10 @@
       return `/${params.map((x) => paramCase(x)).join('/')}`;
     }
 
+    public isExtendedActive(page?: string) {
```
denyo, line 65: This should always be defined since it is rendered in L22. Also if it wasn't defined, then `paramCase` would fail.
```suggestion
    public isExtendedActive(page: string) {
```

denyo (approved):

PR #2992, opened 2024-01-15
Carousel improvements

stephanschroeter:
…l | sas

### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/housekeeping/improve-carousel/components/carousel

#### Scope

- `slot="header"` of Carousel shall always be placed left aligned independent of `alignHeader` prop
- rename `slot="header"` to `slot="controls"`

#### Resolved Issue

Resolves [IssueNumber]


packages/components/src/components/carousel/carousel.tsx
```diff
@@ -245,7 +245,7 @@ export class Carousel {
     warnIfDeprecatedPropIsUsed<typeof Carousel>(this, 'disablePagination', 'Please use pagination prop instead.');
     const hasHeadingPropOrSlot = hasHeading(this.host, this.heading);
     const hasDescriptionPropOrSlot = hasDescription(this.host, this.description);
-    const hasHeaderSlot = hasNamedSlot(this.host, 'header');
+    const hasControlsSlot = hasNamedSlot(this.host, 'controls');
```
denyo, line 248: Slot name could be shared via const for the 3 usages.
What if slot is rendered initially and then removed similar to flyout/modal footer. I guess this change goes undetected?

stephanschroeter: I've abstracted named slots. 
Conditional rendering will be evaluated in a follow up issue.

denyo (approved):

Issue #2916, opened 2023-12-04
[BUG] - `p-modal` / (`p-flyout`?) has missing box-shadow on sticky footer when rendered later


denyo:
### Contact Details
https://porschedev.slack.com/archives/CT7AVHTTQ/p1701449059474459

### Porsche Design System version

latest

### Reproduction Example

https://stackblitz.com/edit/react-vvvjic?file=App.tsx

### Steps to reproduce the behavior

1. Go to https://stackblitz.com/edit/react-vvvjic?file=App.tsx
2. Open example in new tab
3. Click on 'Open Modal'
4. Wait for loading to be finished
5. See missing box-shadow on sticky footer
<img width="377" alt="Bildschirmfoto 2023-12-04 um 14 57 28" src="https://github.com/porsche-design-system/porsche-design-system/assets/4380295/9c392529-4056-4a3e-9aa2-9a72ca31b740">

### Expected behavior

Sticky footer (and also header) box-shadow is correctly shown initially and with later rendered content.  
Probably using `IntersectionObserver` (check git history) is the way to go, since `scroll` event handler is unreliable and not needed anymore since `transform: scale` on open/close was removed with the integration of motion tokens.

### Notes
- Check if issue also appears for Flyout and if it's solvable with  `IntersectionObserver` too.
- We should choose one solution (if possible) which works for both, Modal and Flyout
- Maybe connected to #2955 
- Check if CSS Only Solution would be suitable

michaelworm:
Error is still present in 3.14.0: https://stackblitz.com/edit/react-vvvjic-eqn11t?file=App.tsx

PR #2998, opened 2024-01-17
Refactor Flyout to use native Dialog element to resolve focus issues (issue/2955)

### LINKED ISSUE #2955: [BUG] - Flyout footer slot stays visible
### Contact Details

_No response_

### JS Framework

All

3.9.0

_No response_

### Browser version

_No response_

### Reproduction Example

https://stackblitz.com/edit/react-zjerva-qcgith?file=App.tsx

1. Go to [https://stackblitz.com/edit/react-zjerva-qcgith?file=App.tsx](https://stackblitz.com/edit/react-zjerva-qcgith?file=App.tsx)
2. Click on "Open Flyout"
3. Click on "Footer Button"
4. See footer still being visible after removal of the slot

### Expected behavior

Footer Slot is not visible after removal of the slot

### Relevant log output

_No response_

michaelworm:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2955/components/flyout

#### Scope

Force update of stencil component when slotchange occurs on every slot. 

#### Resolved Issue

Resolves #2955 
Partly fixes issues in #2916


packages/components-vue/src/pages/OverviewPage.vue
```diff
@@ -355,35 +345,35 @@
...
           />
         </PButtonTile>
       </div>
 
       <div class="playground light" title="should render default link-tile-model-signature">
         <!-- eslint-disable vue/no-deprecated-slot-attribute -->
-        <PLinkTileModelSignature :heading="'Some Heading'" :description="'Default'">
+        <PLinkTileModelSignature :description="'Default'" :heading="'Some Heading'">
           <img
+            alt="Beach"
```
denyo, line 373: Is this alphabetical sorting of attributes helpful? I think so far everything was ordered by importance/relevance.

packages/components/src/components/flyout/flyout-styles.ts
```diff
@@ -34,53 +34,73 @@ export const getComponentCss = (
...
           // needed for correct alignment of the Porsche Grid within the Flyout
           '--pds-internal-grid-outer-column': `calc(${spacingFluidLarge} - ${gridGap})`,
           '--pds-internal-grid-margin': `calc(${spacingFluidLarge} * -1)`,
-          justifyContent: isPositionStart ? 'flex-start' : 'flex-end',
           ...getBackdropJssStyle(isOpen, FLYOUT_Z_INDEX, theme),
           ...colorSchemeStyles,
           ...hostHiddenStyles,
         }),
       },
+      dialog: {
```
denyo, line 47: Aren't this identical to flyout-navigation? If yes, why not extract and share them?

packages/components/src/components/flyout/flyout-styles.ts
```diff
@@ -34,53 +34,73 @@ export const getComponentCss = (
...
+        position: 'fixed',
+        'inset-inline': isPositionStart ? '0 0' : 'auto 0',
+        'inset-block': '0 0',
```
denyo, line 50: Is there no `camelCase` version for these css properties available or why do we switch to `'param-case'`?

packages/components/src/components/flyout/flyout.tsx
```diff
@@ -68,56 +66,47 @@ export class Flyout {
   @Watch('open')
   public openChangeHandler(isOpen: boolean): void {
     setScrollLock(isOpen);
-    this.updateFocusTrap(isOpen);
 
     if (isOpen) {
       if (this.hasSubFooter) {
```
denyo, line 71: Could combine conditions with logical AND.

packages/components/src/components/flyout/flyout.tsx
```diff
@@ -68,56 +66,47 @@ export class Flyout {
...
 
   public componentDidRender(): void {
+    // showModal needs to be called after render cycle to prepare visibility states of dialog in order to focus the dismiss button correctly
+    this.setDialogVisibility(this.open);
+
     // TODO: should this really be executed on every rerender, e.g. prop change?
     if (this.open) {
       // TODO: why not scroll to top when opened just like modal does?
 
       if (this.hasSubFooter) {
```
denyo, line 103: Same here.

packages/components/src/components/flyout/flyout.tsx
```diff
@@ -154,31 +143,28 @@ export class Flyout {
...
 
     return (
-      <Host onMouseDown={this.onMouseDown}>
+      <dialog
+        inert={this.open ? null : true} // prevents focusable elements during fade-out transition
+        tabIndex={-1} // dialog always has a dismiss button to be focused
+        ref={(ref) => (this.dialog = ref)}
+        onCancel={this.onCancelDialog}
+        onClick={this.onClickDialog}
+        {...parseAndGetAriaAttributes(this.aria)}
```
denyo, line 152: Is this still fine here or should it go onto `.wrapper`?

packages/components/src/components/flyout/flyout.tsx
```diff
@@ -189,51 +175,78 @@ export class Flyout {
...
+    // prevent closing the dialog uncontrolled by ESC (only relevant for browsers supporting <dialog/>)
+    e.preventDefault();
+
+    this.dismissDialog();
+  };
+
+  private setDialogVisibility(isOpen: boolean): void {
+    if (!this.dialog) {
+      return;
+    }
```
denyo, line 238: We shouldn't need this especially when the reason for adding this is a failing unit test.

packages/components/src/components/modal/modal.tsx
```diff
@@ -79,6 +79,22 @@ export class Modal {
   private hasFooter: boolean;
   private footer: HTMLElement;
   private dialog: HTMLElement;
+  // eslint-disable-next-line @typescript-eslint/member-ordering
```
denyo, line 82: Probably obsolete since it is now next to the other private ones?

packages/components/src/utils/dom/getShadowRootHTMLElements.ts
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable prefer-arrow/prefer-arrow-functions */
+import {getHTMLElements} from './getHTMLElements';
```
denyo, line 2: Looks like this isn't formatted with prettier.
Also what about unit tests?

packages/components/src/utils/focusTrap.ts
```diff
@@ -42,26 +47,42 @@ export const setFocusTrap = (
   let focusableElements: FirstAndLastFocusableElement = [];
 
   document.removeEventListener('keydown', documentKeydownListener);
+
   if (isOpen) {
-    focusableElements = getFirstAndLastFocusableElement(host, closeBtn);
+    // the filter removes falsy values: https://stackoverflow.com/a/41346932
+    focusableElements = getFirstAndLastFocusableElement(host, closeBtn).filter((x) => x);
```
denyo, line 53: Not sure if we ever had a problem with this but if yes, then unit test covering/reproducing this is missing.
I personally would revert all these changes, though, since all the early returns make it more difficult to read and you have to keep all the opposites, e.g. `key === 'Tab'` in mind to understand what is going on. This was clearer before and could have been extended with comments if necessary.

michaelworm: I reverted it

packages/components-vue/src/pages/OverviewPage.vue
michaelworm: It was caused by a Webstorm Setting, I will revert it and disable the sorting

packages/components/src/utils/focusTrap.ts
michaelworm: The refactoring helped me understand it since it was nested if/else before which is, at least for my opinion, harder to understand but the refactoring adds no business value and we want to get rid of the focusTrap at all in the future so I will revert it.

packages/components/src/utils/dom/getShadowRootHTMLElements.ts
michaelworm: I formatted it and added a test for getHTMLElements

packages/components/src/components/flyout/flyout.spec.ts
```diff
@@ -24,20 +23,8 @@ describe('componentDidLoad', () => {
...
-    component.componentDidLoad();
-
-    expect(utilsSpy).toBeCalledWith(
-      component.host,
-      true,
-      component['dialog'],
-      component['dismissBtn'],
-      component['dismissFlyout']
-    );
+    // @ts-ignore
```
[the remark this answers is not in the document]
michaelworm: It was the simpler solution than creating a NodeList just to satisfy the typing error: 

https://stackoverflow.com/a/68023339

Bildschirmfoto 2024-02-20 um 10 31 42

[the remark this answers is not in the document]
michaelworm: I created a NodeList via JS which resolves the typing error

```js
const slotElement = document.createElement('slot');
const slotNodeList = document
      .createElement('div')
      .appendChild(slotElement)
      .appendChild(slotElement)
      .querySelectorAll('slot');
```

packages/components/src/components/flyout/flyout.tsx
michaelworm: In my opinion this is the right place for the aria attributes

packages/components/src/components/flyout/flyout-styles.ts
michaelworm: Done

packages/components-js/tests/e2e/puppeteer/specs/select-wrapper-filter.e2e.ts
```diff
@@ -12,8 +12,7 @@ import {
   setProperty,
   waitForStencilLifecycle,
 } from '../helpers';
-import type { Page } from 'puppeteer';
-import { devices } from 'puppeteer';
+import { KnownDevices, Page } from 'puppeteer';
```
denyo, line 15:
```suggestion
import { KnownDevices, type Page } from 'puppeteer';
```

packages/components-js/tests/unit/fixtures/stats.json
```diff
@@ -1,272 +1,272 @@
 [
   {
     "chunkShortName": "accordion",
-    "size": 4543,
-    "gzipSize": 1980
+    "size": 4919,
```
denyo, line 4: Still a mystery why several components grew quite significantly up too 600 bytes.

packages/components/src/styles/flyout-dialog-reset-styles.ts
```diff
@@ -0,0 +1,14 @@
+import type { Styles } from 'jss';
+
+export const getFlyoutDialogResetStyles = (): Styles => {
```
denyo, line 3: Wrong type?  I think this is a `JssStyle` and function named should be renamed accordingly.

packages/components-js/tests/unit/fixtures/stats.json
henrijoss: We checked the generated chunks and stencil added keys to the slots which correlated to the increase in size

packages/storefront/tests/e2e/specs/accessibility.a11y.ts
```diff
@@ -71,19 +71,16 @@ describe('within main element', () => {
     const buttons = (
       await Promise.all([
         // page.$x("//button[text() = 'Vanilla JS']"),
-        page.$x("//button[text() = 'Angular']"),
-        page.$x("//button[text() = 'React']"),
+        page.waitForSelector("xpath///button[text() = 'Angular']"),
+        page.waitForSelector("xpath///button[text() = 'React']"),
```
denyo, line 75: Isn't this something different now? In the past we got an array of `ElementHandle`s now we just get the first/single one?

packages/storefront/tests/e2e/specs/navigation.e2e.ts
```diff
@@ -130,10 +129,10 @@ it.each<[string, string, string, string | undefined, boolean]>(
   expect(getConsoleErrorsAmount(), `Errors on ${category}/${page}`).toBe(0);
 
   if (tab) {
-    const tabHref = `\/${paramCase(category)}\/${paramCase(page)}\/${paramCase(tab)}`;
-    const [tabElement] = (await browserPage.$x(
-      `//p-tabs-bar//a[contains(., '${tab}')][@href='${tabHref}']`
-    )) as ElementHandle<HTMLElement>[];
+    const tabHref = `/${paramCase(category)}/${paramCase(page)}/${paramCase(tab)}`;
```
denyo, line 132: Escaped slashes got lost, might have an effect on xpath expression.

henrijoss (approved):

henrijoss (dismissed):

packages/components-js/tests/e2e/puppeteer/specs/link-tile-product.e2e.ts
```diff
@@ -1,6 +1,6 @@
 import { expectA11yToMatchSnapshot, getHTMLAttributes, selectNode, setContentWithDesignSystem } from '../helpers';
 import type { Page } from 'puppeteer';
-import { Components } from '@porsche-design-system/components';
+import type { Components } from '@porsche-design-system/components/src/components';
```
stephanschroeter, line 3: Why do we need a `src` cross import instead of a package import?

packages/shared/package.json
```diff
@@ -44,7 +44,8 @@
     "jss": "^10.10.0",
     "jss-plugin-sort-css-media-queries": "^1.0.1-beta.0",
     "jss-preset-default": "^10.10.0",
-    "puppeteer": "~21.6.1",
+    "pixelmatch": "^5.3.0",
```
stephanschroeter, line 47: I think we don't need pixelmatch anymore because all VRT tests are executed by Playwright?

stephanschroeter (dismissed):

packages/components-js/tests/e2e/puppeteer/specs/link-tile-product.e2e.ts
michaelworm: This is needed so that the tests can be run without building the packages

packages/shared/package.json
michaelworm: You are right, this must have been a merge issue. I will remove it.

stephanschroeter (approved):

henrijoss (approved):

stephanschroeter (approved):

PR #2999, opened 2024-01-17
Add CSS Validation Tests

henrijoss:
### Pull Request Checklist

#### References

#### Scope

#### Resolved Issue

Resolves [IssueNumber]


packages/components/tests/unit/helpers/validateCssAndMatchSnapshot.ts
```diff
@@ -2,16 +2,24 @@ import { expect } from '@jest/globals';
...
   TODO: Existing tests for:
-    - !important on host display style
-    - !important on slotted styles
     - @hover media query for hover styles
-    - Host hidden styles
     should also be covered for all getComponentCss conditions
 */
-export const validateCssAndMatchSnapshot = (css: string) => {
-  // We shouldn't use visibility: visible since it cannot be overwritten, use inherit instead
+export const validateCssAndMatchSnapshot = (css: string, isInternal = false) => {
```
denyo, line 8: Could remove the `isInternal` parameter and extract the name of the test suite and/or file tested via `expect.getState()` to get the component name and then retrieve the `isInternal` information via `componentMeta`. This way there is no maintenance and no checking needed what the parameter does.

packages/components/tests/unit/helpers/validateCssAndMatchSnapshot.ts
```diff
@@ -2,16 +2,24 @@ import { expect } from '@jest/globals';
...
+  // We shouldn't use visibility: visible since it cannot be overridden, use inherit instead
   expect(css).not.toMatch(/visibility:\s*(?:var\(.+,\s*visible\s*\)|visible);/);
   // Invalid css which was produced before
   expect(css).not.toMatch('. {');
+  // Expect no !important rule on display style of :host selector since it should be overridable (Match all display with following !important rule)
+  expect(css).not.toMatch(/:host\s*{\s*[^}]*display:\s*(?:var\(\s*--[\w-]+\s*,\s*[\w-]+\s*\)|[\w-]+)\s*!important/);
```
denyo, line 14: Since `\w` is short for `[A-Za-z0-9_]` I guess this matches characters that will never occur, therefore using `[a-z-]` or `[A-Za-z-]` if capital characters are relevant is more precise.  
Similar story for `\s` which matches tabs and line breaks while the only thing that matters are spaces?
`*` could also be replaced with `?` for the most part when there is a single optional space. Or no quantifier at all, when it is always there, e.g. within `display: xyz`.

packages/components/tests/unit/helpers/validateCssAndMatchSnapshot.ts
```diff
@@ -1,17 +1,108 @@
...
 export const validateCssAndMatchSnapshot = (css: string) => {
-  // We shouldn't use visibility: visible since it cannot be overwritten, use inherit instead
-  expect(css).not.toMatch(/visibility:\s*(?:var\(.+,\s*visible\s*\)|visible);/);
-  // Invalid css which was produced before
-  expect(css).not.toMatch('. {');
+  const cssObject = getCssObject(css);
+  const componentName = expect.getState().testPath.match(/\/([^\/]+)\/[^\/]+\.spec\.ts/)[1];
+  const componentTagName = `p-${componentName}` as TagName;
+  // Extract componentMeta from testPath, if it's a functional component this will be undefined
+  const currentComponentMeta = componentMeta[componentTagName];
```
denyo, line 11: I think everywhere else we use the utility function `getComponentMeta` with returns the requested part of the `componentMeta` object, the object itself is just exported for tests.

packages/components/tests/unit/helpers/validateCssAndMatchSnapshot.ts
```diff
@@ -1,17 +1,108 @@
...
+    validateHostDisplayStyle(cssObject);
+  }
+
   expect(css).toMatchSnapshot();
 };
+
+// We shouldn't use visibility: visible since it cannot be overridden, use inherit instead
+const validateVisibilityStyle = (cssObject: object) => {
+  recursivelyApply(cssObject, (key, value) => {
+    if (key.includes('visibility')) {
```
denyo, line 31: Shouldn't this be checked for equality or what could `visibility` be prefixed or suffixed with?

packages/components/tests/unit/helpers/validateCssAndMatchSnapshot.ts
```diff
@@ -1,17 +1,108 @@
...
+  }
+};
+
+// Expect all slotted styles to be !important since they shouldn't be overridable
+const validateSlottedStyles = (cssObject: object, tagName: TagName) => {
+  recursivelyApplyForKey(cssObject, '::slotted', (_, value) => {
+    Object.entries(value).forEach(([cssProp, cssValue]) => {
+      // exceptions for tagName and css property are defined here
+      if (tagName !== 'p-textarea-wrapper' || !['height', 'min-height', 'resize'].includes(cssProp)) {
+        expect(cssValue).toMatch(/ !important$/);
```
denyo, line 53: Maybe remove leading space to discover not using the `addImportantToRule` utility.

packages/components/tests/unit/helpers/validateCssAndMatchSnapshot.ts
```diff
@@ -1,17 +1,108 @@
...
+    });
+  });
+};
+
+// All :hover styles should be wrapped in a @media(hover:hover) since they are only needed for devices which support hover
+const validateHoverMediaQuery = (cssObject: object) => {
+  Object.entries(cssObject).forEach(([key, value]) => {
+    // potential media query
+    if (typeof value === 'object') {
+      Object.entries(value).forEach(([childKey]) => {
```
denyo, line 64: If only keys are relevant, you could use `Object.keys`.

packages/components/tests/unit/helpers/validateCssAndMatchSnapshot.ts
```diff
@@ -1,17 +1,108 @@
...
+    }
+  });
+};
+
+/**
+ * Recursively applies a function to each key-value pair in a nested object.
+ * @param jsonObject - The object to traverse recursively.
+ * @param fn - The function to apply to each key-value pair.
+ */
+const recursivelyApply = (jsonObject: object, fn: (key: string, value: any) => void) => {
```
denyo, line 84: I guess we should rename `jsonObject` since the name makes no sense.

packages/components/tests/unit/helpers/validateCssAndMatchSnapshot.ts
```diff
@@ -1,17 +1,108 @@
...
+
+/**
+ * Recursively applies a function to key-value pairs in a nested object based on a key selector.
+ * @param jsonObject - The object to traverse recursively.
+ * @param keySelector - The key or pattern to match for applying the function.
+ * @param fn - The function to apply to matching key-value pairs.
+ */
+const recursivelyApplyForKey = (jsonObject: object, keySelector: string, fn: (key: string, value: any) => void) => {
+  for (const [key, value] of Object.entries(jsonObject)) {
+    if (key.includes(keySelector)) {
```
denyo, line 102: Function name implies strict equality but implementation is a contains and jsdoc suggest it is a regex.

packages/components/tests/unit/helpers/validateCssAndMatchSnapshot.ts
```diff
@@ -1,17 +1,108 @@
 import { expect } from '@jest/globals';
+import { componentMeta } from '@porsche-design-system/component-meta';
+import { getCssObject } from '../../../src/test-utils';
+import { TagName } from '@porsche-design-system/shared';
```
denyo, line 4:
```suggestion
import type { TagName } from '@porsche-design-system/shared';
```

denyo (approved):

Issue #2931, opened 2023-12-11
[BUG] - Carousel does not work with single pointer event on smaller touch devices


o-hook:
### Description

If the carousel is used on a smaller touch device where the arrow buttons for navigation has been disappeared, it is only possible to navigate to next/prev items by using the swipe mechanism. That violates against WCAG success criterion "[2.5.1 Pointer Gestures](https://www.w3.org/WAI/WCAG22/quickref/#pointer-gestures)" and "[2.5.7 Dragging Movements](https://www.w3.org/WAI/WCAG22/quickref/#dragging-movements)".

### JS Framework

all

### Porsche Design System version

3.9.0

### What browsers are you seeing the problem on?

all

### Browser version

all

### Steps to reproduce the behavior

1. Go to [Carousel](https://designsystem.porsche.com/v3/components/carousel/examples) component with mobile device
2. Try to go to next/previous item without swiping (just by single touch)
3. See error (it's not possible)

### Expected behavior

All functionality that uses multipoint or path-based gestures for operation MUST be operable with a single pointer without a path-based gesture, unless a multipoint or path-based gesture is essential.

- [x] Solution: add possibility to navigate by pagination via click through carousel on mobile and desktop (pagination dots have to be at least 24x24 px)
  - infinite pagination case: consider pagination dots as next/prev buttons next to the centered one
  - none infinite pagination case: each pagination dot represents a specific slide 

### Additional improvements

- [x] `slot="header"` of Carousel shall always be placed left aligned independent of `alignHeader` prop
- [x] rename `slot="header"` to `slot="controls"`
- [x] evaluate why heading is needed to fullfil a11y compliance (remove or change validation if possible, so that heading is not needed to be set) - Validation can be removed and aria prop is recommended in docs when no heading is used.
- [x] add css variable `var(--p-carousel-prev-next-filter, none)` and document on Storefront which enables to invert the theme for prev next buttons by CSS filter value, e.g. `invert(1)` (https://porschedev.slack.com/archives/CT7AVHTTQ/p1704885069366889)

michaelworm:
Currently blocked after discussing in the team by @stephanschroeter

PR #3003, opened 2024-01-18
[BUG] - Carousel does not work with single pointer event on smaller touch devices (issue/2931)

henrijoss:
…s | hj | #2931

### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2931

#### Scope

Adds possibility to navigate by pagination via click through carousel on mobile and desktop. Pagination spacing increases on mobile.

#### Resolved Issue

Resolves [#2931](https://github.com/porsche-design-system/porsche-design-system/issues/2931)


packages/components/src/components/carousel/carousel-styles.ts
```diff
@@ -212,19 +228,36 @@ export const getComponentCss = (
         })),
         position: 'relative',
         justifyContent: isInfinitePagination ? 'flex-start' : 'center',
-        width: `calc(${paginationActiveBulletSize} + ${paginationBulletSize} * 4 + ${spacingStaticSmall} * 4)`, // Width for five bullets (one active + spacing)
-        left: 'calc(50% - 42px)',
+        width: paginationWidth,
+        left: `calc(50% - (${paginationWidth}) / 2)`,
+        ['@media (pointer: coarse)']: {
```
denyo, line 233: Maybe extract this media query into a const with a speaking name since it appears like 3 times.

packages/components/src/components/carousel/carousel-styles.ts
```diff
@@ -212,19 +228,36 @@ export const getComponentCss = (
         })),
         position: 'relative',
         justifyContent: isInfinitePagination ? 'flex-start' : 'center',
-        width: `calc(${paginationActiveBulletSize} + ${paginationBulletSize} * 4 + ${spacingStaticSmall} * 4)`, // Width for five bullets (one active + spacing)
-        left: 'calc(50% - 42px)',
+        width: paginationWidth,
+        left: `calc(50% - (${paginationWidth}) / 2)`,
```
denyo, line 232: Brackets shouldn't be needed I think.
```suggestion
        left: `calc(50% - ${paginationWidth} / 2)`,
```

packages/components/src/components/carousel/carousel.tsx
```diff
@@ -350,7 +348,7 @@ export class Carousel {
         </div>
 
         {(this.disablePagination ? this.disablePagination !== true : this.pagination) && this.hasNavigation && (
-          <div class="pagination-container">
+          <div class="pagination-container" aria-hidden="true">
```
denyo, line 351: Isn't this relevant for a11y? Perhaps using the out of the box pagination of splide would make sense now: https://splidejs.com/guides/pagination/
Or at least use it as an example for what the aria labels are in an accessible scenario on each bullet..

henrijoss: We decided that the pagination is a purely visible feature (like the x-button in the text-field). When navigating with the keyboard you would have to go through all of the slides before reaching the pagination.

packages/components/src/components/carousel/carousel-utils.ts
```diff
@@ -113,14 +108,27 @@ export const isInfinitePagination = (amountOfPages: number): boolean => {
...
     // sanitize in case of removal of slide since activeIndex is from before splide.refresh()
     activeIndex = activeIndex > amountOfPages - 1 ? amountOfPages - 1 : activeIndex;
     paginationEl.innerHTML = Array.from(
       Array(amountOfPages),
       (_, i) => `<span class="bullet${i === activeIndex ? ' ' + bulletActiveClass : ''}"></span>`
     ).join('');
+
+    const bullets = paginationEl.querySelectorAll('.bullet');
+    bullets.forEach((bullet, i) => {
+      bullet.addEventListener('click', () => {
```
denyo, line 127: Would be better to have single click listener on the parent and then figure out which bullet was clicked in the handler.

PR #3009, opened 2024-01-24
Optimize loading state behavior for assistive technologies (Issue/2495)

### LINKED ISSUE #2495: Optimize loading state behavior for assistive technologies
Scope
---
The PDS has several components which can have an "intermediate" `loading` state (while in focus). Those components must expose this state to assistive technology users (e.g. screen reader) including telling the user when loading starts and when it ends.

Out of Scope
---
- No internationalisation, predefined English texts only

Notes  
---
- See open Polaris issue as inspiration: https://github.com/Shopify/polaris/issues/1212
- Article about an accessible microinteraction button: https://accessabilly.com/an-accessible-microinteractions-button/
- Article about multi-functional buttons: https://adrianroselli.com/2021/01/multi-function-button.html
- Handling of the  begin/end loading state should be equal for those components

Acceptance criteria
---
- [ ] All components with loading states should expose begin/end of this state to assistive technology users

Subtasks
---
- [ ] Task

o-hook:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/2495/

#### Scope

Optimize loading state behavior for assistive technologies

#### Resolved Issue

Resolves #2495


packages/components/src/components/common/loading-message/loading-message-styles.ts
```diff
@@ -0,0 +1,8 @@
+import { type Styles } from 'jss';
+import { getHiddenTextJssStyle } from '../../../styles';
+
+export const getFunctionalComponentLoadingMessageStyles = (): Styles<'status'> => {
+  return {
+    status: getHiddenTextJssStyle(),
```
stephanschroeter, line 6: I think the style class should be named like the functional component is named to be in sync with other functional components.

packages/components-js/tests/unit/fixtures/stats.json
```diff
@@ -11,8 +11,8 @@
   },
   {
     "chunkShortName": "button",
-    "size": 7229,
-    "gzipSize": 2894
+    "size": 7696,
```
stephanschroeter, line 14: Sizing increased "heavily" for affected components, but why?

o-hook: I assume that this is caused by our `getHiddenTextJssStyle()` in combination with the new HTML node. Is there a possibility to extract this "global" styles to be rendered just once and not every time when it's used?

packages/components/src/components/checkbox-wrapper/checkbox-wrapper.tsx
```diff
@@ -64,6 +66,13 @@ export class CheckboxWrapper {
     }
   }
 
+  @Watch('loading')
```
denyo, line 69: This won't cover the initial state when the component is rendered with `loading=true` which is the removed afterwards. 
Is this intentional? 

Bildschirmfoto 2024-01-29 um 08 55 46


Here I would expect `Loading finished` but instead it is empty.
Bildschirmfoto 2024-01-29 um 08 55 53

packages/components/src/components/button/button.tsx
```diff
@@ -80,13 +81,22 @@ export class Button {
...
+
   @Listen('click', { capture: true })
   public onClick(e: MouseEvent): void {
     if (isDisabledOrLoading(this.disabled, this.loading)) {
       e.stopPropagation();
     }
   }
 
+  @Watch('loading')
+  public loadingChanged(newVal: boolean): void {
```
denyo, line 94: It would be worth looking into decorators to share this across components. Inheritance would also work but I guess isn't supported.
Also a generic test based on `component-meta` to establish this rule for all present and future components with the `loading` prop would be helpful.

packages/components/src/components/checkbox-wrapper/checkbox-wrapper.tsx
o-hook: Good catch! Should be fixed now.

packages/components/src/components/button/button.tsx
o-hook: Do you mean abstracting the @Watch decorator by using it inside of a custom decorator which then can be called with smth. like this? `@CustomLoadingDecorator`

packages/components/src/components/switch/switch-styles.ts
```diff
@@ -213,5 +214,6 @@ export const getComponentCss = (
         )
       ),
     },
+    ...getFunctionalComponentLoadingMessageStyles(),
```
denyo, line 217:
```suggestion
    // .loading
    ...getFunctionalComponentLoadingMessageStyles(),
```

packages/components/src/components/button-pure/button-pure-styles.ts
```diff
@@ -53,6 +54,7 @@ export const getComponentCss = (
               height: fontLineHeight,
             },
           }),
+        ...getFunctionalComponentLoadingMessageStyles(),
```
denyo, line 57:
```suggestion
        // .loading
        ...getFunctionalComponentLoadingMessageStyles(),
```

denyo (approved):

PR #3055, opened 2024-02-21
[BUG] - Carousel: "skip-entries" link not visible on :focus-visible (Issue/3050)

### LINKED ISSUE #3050: [BUG] - Carousel: "skip-entries" link not visible on `:focus-visible`
### JS Framework

All

3.12.0

### Reproduction Example

https://stackblitz.com/edit/react-gitozp?file=App.tsx

1. Go to [Stackblitz](https://stackblitz.com/edit/react-gitozp?file=App.tsx)
2. Press Tab-Key until the prev/next buttons gets focus.
3. See error: the tab before the "prev" button focuses the "skip-entries" link but it is still hidden

### Expected behavior

The "skip-entries" link should become visible on `:focus-visible` 

### Notes

- Somehow `:focus-visible` is not getting applied on `.skip-link` although custom element should be the `activeElement`.
- Seams like Skip Link behaviour is not yet covered by test automation?

o-hook:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3050/

#### Scope

#### Resolved Issue

Resolves #3050

### DIFF
--- a/packages/components-angular/tests/vrt/specs/styles/focus.vrt.ts
+++ b/packages/components-angular/tests/vrt/specs/styles/focus.vrt.ts
@@ -1,5 +1,5 @@
 import { expect, test } from '@playwright/test';
-import { forceFocusState } from '../../../../../components-js/tests/vrt/playwright/helpers';
+import { forceFocusState, forceFocusVisibleState } from '../../../../../components-js/tests/vrt/playwright/helpers';
 import { viewportWidthM } from '@porsche-design-system/shared/testing/playwright.vrt.config';
 
 const style = 'styles-focus';
@@ -9,7 +9,9 @@ test.describe(style, async () => {
     await page.goto(`/${style}`);
     await page.setViewportSize({ width: viewportWidthM, height: 600 });
     await forceFocusState(page, 'a');
+    await forceFocusVisibleState(page, 'a');
     await forceFocusState(page, 'button');
+    await forceFocusVisibleState(page, 'button');
     await expect(page.locator('#app')).toHaveScreenshot(`${style}-${viewportWidthM}.png`);
   });
 });
--- a/packages/components-react/tests/vrt/specs/styles/focus.vrt.ts
+++ b/packages/components-react/tests/vrt/specs/styles/focus.vrt.ts
@@ -1,5 +1,5 @@
 import { expect, test } from '@playwright/test';
-import { forceFocusState } from '../../../../../components-js/tests/vrt/playwright/helpers';
+import { forceFocusState, forceFocusVisibleState } from '../../../../../components-js/tests/vrt/playwright/helpers';
 import { viewportWidthM } from '@porsche-design-system/shared/testing/playwright.vrt.config';
 
 const style = 'styles-focus';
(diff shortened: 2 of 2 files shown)


packages/components/src/components/carousel/carousel-styles.ts
```diff
@@ -178,7 +178,8 @@ export const getComponentCss = (
     'skip-link': {
       opacity: 0,
       pointerEvents: 'none',
-      '&:focus-visible': {
+      // :focus mut be used in this case, because :focus-visible is just matched on the focusable element itself, not on the host element.
```
henrijoss, line 181:
```suggestion
      // :focus must be used in this case, because :focus-visible is just matched on the focusable element itself, not on the host element.
```

packages/components-js/tests/e2e/puppeteer/specs/carousel.e2e.ts
```diff
@@ -1141,4 +1142,15 @@ describe('accessibility', () => {
...
   });
+
+  it('should change skip-link to visible if it receives keyboard focus', async () => {
+    await initCarousel({ skipLinkTarget: '#link-after' });
+    const skipLinkHost = await getSkipLinkHost();
+
+    expect(await getElementStyle(skipLinkHost, 'opacity')).toBe('0');
+
+    await page.keyboard.press('Tab');
+
```
henrijoss, line 1153: Maybe also check that the focus is on the skip link

o-hook: @henrijoss  Isn't this indirectly checked with the CSS selector `:focus`? Because this selector controls the opacity which then gets checked by the test.

packages/components/src/components/carousel/carousel-styles.ts
```diff
@@ -178,7 +178,8 @@ export const getComponentCss = (
     'skip-link': {
       opacity: 0,
       pointerEvents: 'none',
-      '&:focus-visible': {
+      // :focus must be used in this case, because :focus-visible is just matched on the focusable element itself, not on the host element.
+      '&:focus': {
         opacity: 1,
         pointerEvents: 'all',
```
denyo, line 184: Couldn't this entire part be defined the other way around without overriding/resetting?
``` 
'skip-link:not(:focus)': {
  opacity: 0,
  pointerEvents: 'none',
} 
```

packages/components-js/tests/e2e/puppeteer/specs/carousel.e2e.ts
denyo: I guess additionally checking the active element wouldn't hurt, since focus isn't explicitly covered. Currently this test would be green by having a keydown listener and adding a class or inline style on the element. 

Besides, what about the states vrts? These wrongly show the skip link button then since selectors are too greedy/wrong?

packages/components/src/components/carousel/carousel-styles.ts
o-hook: Hey @denyo : 
Do you know why the following code works but the latter not?

```
'skip-link': {
      '&:not(:focus)': {
        opacity: 0,
        pointerEvents: 'none',
      },
    },
```


```
'skip-link:not(:focus)': {
      opacity: 0,
      pointerEvents: 'none',
    },
```

packages/components-react/projects/react-ssr-wrapper/tests/unit/specs/__snapshots__/react-ssr-wrapper.spec.tsx.snap
```diff
@@ -6771,10 +6771,6 @@ exports[`should render dsr component for PCarousel 1`] = `
 .btn {
   padding: 8px;
 }
-.skip-link {
```
stephanschroeter, line 6774: is it correct?

stephanschroeter (approved):

PR #3068, opened 2024-02-27
[Feature Request | Bug] Text Field Wrapper with <input type="date|time" /> shall visualize ::placeholder state in Safari browser

### LINKED ISSUE #3043: [Feature Request | Bug]  Text Field Wrapper with `<input type="date|time" />` shall visualize `::placeholder` state in Safari browser
Scope
---
When using the Text Field Wrapper with `<input type="date|time" />` in Safari then the browser shows a placeholder by default. Unfortunately, the placeholder value is coloured like an actually entered value. This might be confusing for a user, therefore we want to properly style the placeholder for Text Field Wrapper.

Notes  
---
- Community Request: https://porschedev.slack.com/archives/CT7AVHTTQ/p1707902763303759
- Possible solutions:
  - https://stackoverflow.com/questions/73866581/how-to-stop-safari-16-macos-defaulting-to-the-current-date-in-a-html-type-dat
  - `::placeholder` is not defined for Text Field Wrapper at all? (`packages/components/src/styles/placeholder.ts`)
  - Don't define `color` and rely on browser user agent style?
  - Maybe `:valid` or `:invalid` CSS selector can help?
- Also have look why date/time on mobile Safari is not aligned correctly (maybe try to enable mobile browser emulation in Playwright)?

Image

Acceptance criteria
---
- [x] Placeholder for Text Field Wrapper is visualised correctly for all `<input>` types

Subtasks
---
- [x] fix appearance of `input type="date"` and `input type="time"` for mobile safari
- [x] try `:valid` or `:invalid` selectors (pointless, since empty value is valid, too)
- [x] try `:placeholder` selector (useless, `input type="date"` and `input type="time"` don't show placeholders)
- [x] try `:after` pseudo element with `content: attr(placeholder)` (works, but needs absolute positioning and be hidden on `:focus` and when value was entered)
- [x] try safari pseudo selectors `input::-webkit-datetime-edit-fields-wrapper, input::-webkit-datetime-edit-day-field, input::-webkit-datetime-edit-month-field, input::-webkit-datetime-edit-year-field` (works only partially since `color` isn't respected, but `filter` is which leads to other issues, also this is applied to actual values, not just the placeholder, too)
- [x] try doing something with `opacity` (similar situation as before as there is no CSS only way to detect if we are in placeholder or value mode which)
- [x] try no dedicated color (only works if no `color` and no `background` is specified on the `slotted(input)` which could probably be applied via safari media query)
- [x] add workaround for native placeholder behavior in safari

denyo:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3043/

#### Scope

- fix alignment of `p-text-field-wrapper` for `type="date"` and `type="time"` on mobile safari
- fix native placeholder color style for safari on `theme="light"`, for `theme="dark"` this isn't possible and would cause `:` or `.` within formatted time or date to disappear and also make the actual value disappear

### Chrome
Bildschirmfoto 2024-03-04 um 14 44 55
Bildschirmfoto 2024-03-04 um 14 45 02

### Safari
Bildschirmfoto 2024-03-04 um 14 44 40
Bildschirmfoto 2024-03-04 um 14 44 46

### Mobile Safari
Simulator Screenshot - iPhone 15 - 2024-03-04 at 14 47 03
Simulator Screenshot - iPhone 15 - 2024-03-04 at 14 46 53

#### Resolved Issue

Resolves #3043


packages/components/src/components/text-field-wrapper/text-field-wrapper-styles.ts
```diff
@@ -63,6 +69,15 @@ export const getComponentCss = (
           ...(isNumber && {
             MozAppearance: 'textfield', // hides up/down spin button for Firefox
           }),
+          ...((isCalendar || isTime) && {
+            // for native placeholder color in safari, background ahs to be a special value, `transparent` or `rgba(0,0,0,0)` won't work
```
stephanschroeter, line 73:
```suggestion
            // for native placeholder color in safari, background has to be a special value, `transparent` or `rgba(0,0,0,0)` won't work
```

packages/components/src/components/text-field-wrapper/text-field-wrapper-styles.ts
```diff
@@ -63,6 +69,15 @@ export const getComponentCss = (
...
+            // this works nice for `theme="light"`, but for `theme="dark"` placeholder appears blue which is still better
+            // than having invisible dots or colons for data/time or not seeing the value at all after selection
+            '@supports (-webkit-hyphens: none)': {
```
stephanschroeter, line 76: What is this media query for? Is it a Safari only media query?

stephanschroeter (approved):

packages/components/src/components/text-field-wrapper/text-field-wrapper-styles.ts
```diff
@@ -63,6 +69,15 @@ export const getComponentCss = (
...
+              background: 'rgba(0,0,1,0)',
+              color: isThemeDark(theme) ? getThemedColors(theme).primaryColor : 'initial',
```
stephanschroeter, line 78: Why can't we set `color: "initial"` always? I thought `theme=dark` has no effect in Safari anyways?

stephanschroeter, line 78: Does the WebKit VRT test show the new behaviour?

packages/components/src/components/text-field-wrapper/text-field-wrapper-styles.ts
denyo: Added the link to the original source.

packages/components/src/components/text-field-wrapper/text-field-wrapper-styles.ts
denyo: No, the safari screenshots don't show any placeholder at all and didn't do so before (which would have been with wrong colors.

Bildschirmfoto 2024-03-05 um 08 28 57

denyo: See native placeholders at the top with black `.` and `:` characters. These won't be visible on black background, only placeholder is. Once a value is chosen/entered the value will disappear as well since it is black by default.

Bildschirmfoto 2024-03-05 um 08 30 40

stephanschroeter (approved):

PR #3181, opened 2024-04-15
Sticky Header for Accordion as `<details>`

### LINKED ISSUE #3036: [Feature Request] Sticky Header for Accordion
Scope
---
Describe clearly the scope of the task.

Out of Scope
---

Notes  
---
Share any further information that is relevant for processing.

Acceptance criteria
---
- [ ] Criteria

Subtasks
---
- [ ] Task

michaelworm:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3036/

#### Scope

#### Resolved Issue

Resolves #3036


packages/components/CHANGELOG.md
```diff
@@ -14,12 +14,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0),
 
 ### [Unreleased]
 
+- `Accordion`: Add property "sticky" for a fixed heading
```
henrijoss, line 17:
```suggestion
- `Accordion`: Add experimental property `sticky` for a fixed heading
```

henrijoss:
Maybe an E2E test would make sense?

packages/components/src/components/accordion/accordion.tsx
```diff
@@ -71,7 +77,7 @@ export class Accordion {
 
     return (
       <Host>
-        <Heading class="heading">
+        <Heading class={{ heading: true, sticky: this.sticky }}>
```
stephanschroeter, line 80: Why do we need a sticky css class?

packages/components/src/components/accordion/accordion-styles.ts
```diff
@@ -115,6 +121,15 @@ export const getComponentCss = (
     },
     heading: {
       margin: 0,
+      ...(sticky && {
+        position: 'sticky',
+        top: 0,
+        zIndex: zIndexes.indexOf('heading'),
```
stephanschroeter, line 127: why not simply define `zIndex: 1` (maybe with a comment) then we'd save unnecessary bytes in the bundled chunk?

packages/components/src/components/accordion/accordion-styles.ts
```diff
@@ -136,6 +151,10 @@ export const getComponentCss = (
         color: primaryColorDark,
       }),
       display: 'grid',
+      ...(sticky && {
+        position: 'relative',
+        zIndex: zIndexes.indexOf('collapsible'),
```
stephanschroeter, line 156: why not simply define `zIndex: 0` (maybe with a comment) then we'd save unnecessary bytes in the bundled chunk?

packages/components/src/components/accordion/accordion-styles.ts
```diff
@@ -28,19 +28,25 @@ export const getComponentCss = (
...
+    backgroundColor: backgroundColorDark,
   } = getThemedColors('dark');
+  const zIndexes = ['collapsible', 'heading'];
 
   return getCss({
     '@global': {
       ':host': {
         display: 'block',
+        ...(sticky && {
+          position: 'relative',
```
stephanschroeter, line 48: a comment would be helpful why a `position` is needed.

packages/components/src/components/accordion/accordion.tsx
michaelworm: You are right, in an earlier stage I used this to style it sticky but then I changed it to pass "sticky" into the jss.

packages/components/src/components/accordion/accordion-styles.ts
michaelworm: Coming from a Sass background I found it way more intuitive to define such z-index ordering as a list, but you are right, we need to primarily safe bytes and a comment should be fine.

packages/components/src/components/accordion/accordion-styles.ts
michaelworm: Coming from a Sass background I found it way more intuitive to define such z-index ordering as a list, but you are right, we need to primarily safe bytes and a comment should be fine.

packages/components/src/components/accordion/accordion-styles.ts
michaelworm: Actually I looked it up and it's not needed since position sticky tells the parent already to be relative, so this is not needed. Thanks!

PR #3191, opened 2024-04-22
Modal goes `<dialog />` (issue/2682)

### LINKED ISSUE #2682: Improve focus trap for `p-modal`
Scope
---
Based on feedback from our community there is a more reliable way to trap the focus within a `p-modal` ~~by  inserting invisible focus elements at the front and back of the focus trap content instead of getting the list of focusable elements within the content each time on slot change and registering tab event handlers for the last and first elements, like the PDS currently does.~~ by using the correct and newly introduced HTML element `<dialog/>` which takes care of focus trap behaviour. 

~~This wouldn't just simply things but also remove the current limitation that we only detect further DOM changes like adding or removing DOM nodes only on the first level, hence direct children of the modal.~~

It would also **fix a bug in screen reader usage** where it's possible to access also elements outside of the modal which is not accessibility compliant.

Notes  
---
- Use `<dialog/>` like in Flyout Navigation then no focus trap is needed
  - should visually still work in browser which are not supporting `<dialog />` (`#top-layer` support vs `position: fixed` as fallback)
  - focus trap is not supported for browsers which are not supporting `<dialog />`
- Custom Focus Trap function can be removed when not used anymore
- https://porschedev.slack.com/archives/CT7AVHTTQ/p1689762069786679
- https://porschedev.slack.com/archives/CT7AVHTTQ/p1705318929655209
- https://porschedev.slack.com/archives/CT7AVHTTQ/p1715179858397279
- https://web.dev/learn/html/dialog
- https://www.stefanjudis.com/blog/a-look-at-the-dialog-elements-super-powers/
- Ensure `ESC` closes the Modal (currently not working 100% for Flyout Navigation)
- TBD.: Is it useful to define a standardized z-index for Cookie Consent Banner, Country Selection Navbar and Live Chat, see: https://designsystem.porsche.com/v3/patterns/notifications/introduction#z-index-in-descending-order ? Maybe, it doesn't even matter anymore with #top-layer and <dialog />, see: https://developer.chrome.com/blog/what-is-the-top-layer
- TBD: Should we use ::backdrop styling for Modal, Flyout and Flyout-Navigation (https://caniuse.com/?search=%3A%3Abackdrop)? Otherwise stacking context mechanism is not guaranteed (like currently).

Acceptance criteria
---
- [ ] `Modal` component is using native `dialog` element to handle focus trap.

Subtasks
---
- [ ] Task

stephanschroeter:
### Pull Request Checklist

#### References

- Preview: 
  - https://designsystem.porsche.com/issue/2682/components/modal
  - https://designsystem.porsche.com/issue/2682/components/flyout

#### Scope

- Modal uses `<dialog />`
- Focus Trap works reliably
- Layout/Spacing is aligned between Flyout and Modal
- Sticky area visualisation is determined with IntersectionObserver instead of ScrollListener

#### Resolved Issue

Resolves #2682


packages/components/src/index.html
```diff
@@ -105,7 +115,188 @@ <h1>Porsche Design System - Playground</h1>
       </select>
     </header>
 
-    <main></main>
+    <main>
```
stephanschroeter, line 118: revert

packages/components/src/components/modal/modal-utils.ts
```diff
@@ -1,10 +1,11 @@
-import { type Backdrop } from '../../styles';
+import { type Backdrop } from '../../styles/dialog-styles';
 
 export const MODAL_ARIA_ATTRIBUTES = ['aria-label'] as const;
 export type ModalAriaAttribute = (typeof MODAL_ARIA_ATTRIBUTES)[number];
 
 export type ModalBackdrop = Backdrop;
 
+// TODO: can be removed?
```
stephanschroeter, line 8: check

henrijoss (approved):

PR #3215, opened 2024-05-08
Update playwright dependency

michaelworm:
### Pull Request Checklist


michaelworm:
Closed for now since Playwright in this version 1.44.0 does not show styled ::before elements in VRT screenshots, which we use for displaying the VRT test case title and also for checkbox checked state. Waiting for the next release.

PR #3220, opened 2024-05-14
[BUG] - Carousel: aria-label is not set if slotted heading is provided (Issue/3064)

### LINKED ISSUE #3064: [BUG] - Carousel: `aria-label` is not set if slotted heading is provided
### Contact Details

_No response_

### JS Framework

all

3.12.0

_No response_

### Browser version

all

### Reproduction Example

...

4. See error'

### Expected behavior

The `aria-label`property should also be added to the `id="splide` element (as it's given by providing the heading through the property)

### Relevant log output

_No response_

o-hook:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3064/

#### Scope

#### Resolved Issue

Resolves #3064

### DIFF
--- a/packages/components-js/tests/a11y/specs/others/carousel.a11y.ts
+++ b/packages/components-js/tests/a11y/specs/others/carousel.a11y.ts
@@ -12,6 +12,8 @@ import type {
 } from '@porsche-design-system/components/dist/types/bundle';
 
 type InitOptions = {
+  heading?: boolean;
+  slottedHeading?: boolean;
   aria?: SelectedAriaAttributes<CarouselAriaAttribute>;
   slidesPerPage?: number | string;
   amountOfSlides?: number;
@@ -23,6 +25,8 @@ type InitOptions = {
 
 const initCarousel = (page: Page, opts?: InitOptions) => {
   const {
+    heading = true,
+    slottedHeading = false,
     aria,
     slidesPerPage = 1,
     amountOfSlides = 3,
@@ -42,6 +46,7 @@ const initCarousel = (page: Page, opts?: InitOptions) => {
   const focusableElementBefore = withFocusableElements ? '<a id="link-before" href="#">Link before</a>' : '';
   const focusableElementAfter = withFocusableElements ? '<a id="link-after" href="#">Link after</a>' : '';
   const attrs = [
+    heading ? `heading="Some heading"` : '',
     aria && `aria="${aria}"`,
     slidesPerPage ? `slides-per-page="${slidesPerPage}"` : '',
     rewind === false ? 'rewind="false"' : '',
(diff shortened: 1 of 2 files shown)


packages/components-js/tests/a11y/specs/others/carousel.a11y.ts
```diff
@@ -146,12 +154,29 @@ test('should expose correct initial accessibility tree', async ({ page }) => {
   expect(await getAttribute(slide3, 'aria-label')).toBe('3 of 3');
 });
 
-test.fixme('should expose correct initial accessibility tree when aria prop is defined', async ({ page }) => {
-  await initCarousel(page, { aria: "{'aria-label': 'Other Heading'}" });
+test('should expose correct aria-label on splide element if slotted heading is set', async ({ page }) => {
```
henrijoss, line 157: test decription doesn't match what is being tested? only tests for `aria-labelledby` and not `aria-label`?

packages/components-js/tests/a11y/specs/others/carousel.a11y.ts
```diff
@@ -146,12 +154,29 @@ test('should expose correct initial accessibility tree', async ({ page }) => {
...
+  await initCarousel(page, { heading: false, slottedHeading: true });
   const splide = await getSplide(page);
 
-  // await expectA11yToMatchSnapshot(page, splide, { message: 'splide', interestingOnly: false });
+  expect(await getAttribute(splide, 'aria-labelledby')).toBe('heading');
+});
+
+test('should expose correct aria-label when aria prop is defined and heading is set', async ({ page }) => {
+  await initCarousel(page, { aria: "{'aria-label': 'Other heading'}" });
```
henrijoss, line 165: Maybe extract the `Other heading` into a const and use in the `expect` below. Also goes for the other test.

packages/components-js/tests/a11y/specs/others/carousel.a11y.ts
henrijoss, line 157: Do we also need a test if `slottedHeading` is true and an `aria-label` is set manually?

packages/components/src/components/carousel/carousel-styles.ts
```diff
@@ -57,7 +57,7 @@ const paginationWidthLarge = `calc(${paginationActiveBulletSize} + ${paginationB
   paginationVisibleBulletCount - 1
 } + ${paginationGapLarge} * ${paginationVisibleBulletCount - 1} + 2 * ${paginationInset})`; // Width for one active bullet + width of inactive bullets + spacing
 
-const selectorHeading = 'h2,::slotted([slot="heading"])';
+const selectorHeading = ':is(.heading)';
```
henrijoss, line 60: Why do we need the `:is` selector here, doesn't selecting `.heading` would be enough?

o-hook: We need the `:is` selector to have the possibilities to combine selectors on `global` space, e.g. in line 121:
`...((hasHeading || hasDescription) && {`

henrijoss: `:is(.heading),p,::slotted([slot="description"])` and `.heading,p,::slotted([slot="description"])` shouldn't make a difference

PR #3307, opened 2024-05-22
Fix banner position on mobile (issue/3129)

### LINKED ISSUE #3129: [FEATURE] Migrate Banner to `<dialog popover />`
Scope
---
We've refactored the Flyout component to make use of `#top-layer` which causes the Banner not to be positioned on top of the Flyout. Therefor, we need to ensure that the Banner is getting rendered on `#top-layer` too.

Notes  
---
- Ensure Banner is rendered on `#top-layer` (which is not the case when `<dialog />` gets opened by `.show()`)
- Consider using attribute `popover` on either a `<div />` or custom element `:host` or `<dialog />` opened by `.show()` in combination with attr `popover`… or maybe CSS `overlay: auto` is good enough
- https://developer.mozilla.org/en-US/docs/Web/CSS/overlay
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog?retiredLocale=de
- https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover

Acceptance criteria
---
- [x] Banner is rendered on `#top-layer`

Subtasks
---
- [ ] Task

michaelworm:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3129/

#### Scope

Since the banner is now a top-layer element, the top-position needs to be reset in order to eliminate the inset style coming from the browser to display the banner on the bottom.

#### Resolved Issue

Resolves #3129

### DIFF
--- a/packages/components/src/components/banner/banner-styles.ts
+++ b/packages/components/src/components/banner/banner-styles.ts
@@ -33,6 +33,7 @@ export const getComponentCss = (isOpen: boolean): string => {
         display: 'block',
         ...addImportantToEachRule({
           ...getBannerPopoverResetStyles(),
+          top: 'auto',
           bottom: `var(${cssVariableBottom},${topBottomFallback})`,
           left: gridExtendedOffsetBase,
           right: gridExtendedOffsetBase,


packages/components/src/components/banner/banner-styles.ts
```diff
@@ -33,12 +33,16 @@ export const getComponentCss = (isOpen: boolean): string => {
         display: 'block',
         ...addImportantToEachRule({
           ...getBannerPopoverResetStyles(),
+          top: 'auto',
```
henrijoss, line 36: Maybe better to overwrite the UA inset style in `getBannerPopoverResetStyles()` e.g. `inset: auto`

stephanschroeter (approved):

PR #3314, opened 2024-05-29
[BUG] Refactor parseJSONAttribute to support Safari < 16.4 (issue/3313)

### LINKED ISSUE #3313: [BUG] - Regex lookbehind in `components/src/utils/json.ts` is not supported in Safari < 16.4
### Contact Details

https://porschedev.slack.com/archives/CT7AVHTTQ/p1716889811670299

### JS Framework

Next.js

3.15.1

iOS/Mac

Safari, Mobile Safari

### Browser version

< 16.4

### Reproduction Example

Download 16.2 iOS Simulator in Xcode and run the Next.js example repo (with 3.15.1 installed).
_Hint: The bug is not related to the SSR environment, it should also be broken when starting components playground._

Add an aria-label to any component, then the error should appear.

In Release 3.15.0 we introduced Regex lookbehinds here: https://github.com/porsche-design-system/porsche-design-system/pull/3217

This is not supported in Safari prior 16.4 => https://caniuse.com/js-regexp-lookbehind which is causing errors for one of our customers (Dealbuilder)

Bildschirmfoto 2024-05-28 um 13 57 33
Bildschirmfoto 2024-05-28 um 14 06 58

### Expected behavior

Rewrite the Regex or find another solution to achieve support of single quotes for Safari < 16.4

### Relevant log output

_No response_

michaelworm:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3313/

#### Scope

By avoiding using regex lookbehinds the function is working again in Safari < 16.4

#### Resolved Issue

Resolves #3313

### DIFF
--- a/packages/components/src/utils/json.ts
+++ b/packages/components/src/utils/json.ts
@@ -4,17 +4,15 @@ export const parseJSONAttribute = <T>(attribute: T | string): T => {
     return attribute;
   }
 
-  // Convert single quotes to double quotes except the ones which are escaped by backslash
-  let jsonString = attribute
-    .replace(/\\'/g, '__escaped_single_quote__')
-    .replace(/'/g, '"')
-    .replace(/__escaped_single_quote__/g, '\\\'');
-
-  // Remove string escapes except the ones followed by unicode u0027
-  jsonString = jsonString.replace(/([^\\])\\(?!u0027)/g, '$1');
-
-  // Wrap keys in double quotes
-  jsonString = jsonString.replace(/[\s"]?([\w-]+)[\s"]?:/g, '"$1":');
-
-  return JSON.parse(jsonString);
+  return JSON.parse(
+    attribute
+      // Convert single quotes to double quotes except the ones which are escaped by backslash
+      .replace(/\\'/g, '__escaped_single_quote__')
+      .replace(/'/g, '"')
+      .replace(/__escaped_single_quote__/g, '\\\'')
+      // Remove string escapes except the ones followed by unicode u0027
+      .replace(/([^\\])\\(?!u0027)/g, '$1')
+      // Wrap keys in double quotes
+      .replace(/[\s"]?([\w-]+)[\s"]?:/g, '"$1":')
+  );
 };
--- a/packages/components/src/utils/json.ts
+++ b/packages/components/src/utils/json.ts
@@ -1,12 +1,20 @@
 export const parseJSONAttribute = <T>(attribute: T | string): T => {
-  return typeof attribute === 'string'
-    ? // input is potentially JSON parsable string, e.g. "{ 'aria-label': 'Some label' }"
-      JSON.parse(
-        attribute
-          .replace(/(?<!\\)'/g, '"') // convert single quotes to double quotes except the ones which are escaped by backslash
-          .replace(/\\(?!u0027)/g, '') // remove string escapes except the ones followed by unicode u0027
-          .replace(/[\s"]?([\w-]+)[\s"]?:/g, '"$1":') // wrap keys in double quotes
-      )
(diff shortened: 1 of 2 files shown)


henrijoss:
Maybe we could add a console error message if the parsing failed

packages/components/src/utils/json.ts
```diff
@@ -1,12 +1,20 @@
...
+  }
+
+  // Convert single quotes to double quotes except the ones which are escaped by backslash
+  let jsonString = attribute
+    .replace(/\\'/g, '__escaped_single_quote__')
+    .replace(/'/g, '"')
+    .replace(/__escaped_single_quote__/g, '\\\'');
+
+  // Remove string escapes except the ones followed by unicode u0027
+  jsonString = jsonString.replace(/([^\\])\\(?!u0027)/g, '$1');
```
stephanschroeter, line 14: why can't we add the `replace` functions to where the others are and get rid of the let variable?

packages/components/src/utils/json.ts
```diff
@@ -1,12 +1,20 @@
...
-        attribute
-          .replace(/(?<!\\)'/g, '"') // convert single quotes to double quotes except the ones which are escaped by backslash
-          .replace(/\\(?!u0027)/g, '') // remove string escapes except the ones followed by unicode u0027
-          .replace(/[\s"]?([\w-]+)[\s"]?:/g, '"$1":') // wrap keys in double quotes
-      )
-    : // input is object, e.g. { 'aria-label': 'Some label' }
-      attribute;
+  // Input is object, e.g. { 'aria-label': 'Some label' }
+  if (typeof attribute !== 'string') {
+    return attribute;
```
stephanschroeter, line 4: Isn't a single return per function preferred over multiple returns?

packages/components/src/utils/json.ts
michaelworm: Readability, but sure we can omit the variable

packages/components/src/utils/json.ts
michaelworm: This adds alot to readability since the ternary function was complex to read and understand. Also I never heard of a single return policy for functions, but feel free to lighten me up :)

packages/components/src/utils/json.ts
michaelworm: Result (harder to read in my opinion)
```ts
return JSON.parse(
    attribute
      // Convert single quotes to double quotes except the ones which are escaped by backslash
      .replace(/\\'/g, '__escaped_single_quote__')
      .replace(/'/g, '"')
      .replace(/__escaped_single_quote__/g, '\\\'')
      // Remove string escapes except the ones followed by unicode u0027
      .replace(/([^\\])\\(?!u0027)/g, '$1')
      // Wrap keys in double quotes
      .replace(/[\s"]?([\w-]+)[\s"]?:/g, '"$1":')
  );
```

stephanschroeter: I don't think it's more readable in this case to be honest. And I think working with `const` is always to be preferred over `let`.

michaelworm: Okay, sure. I will push the proposed version above.

packages/components/src/utils/json.ts
stephanschroeter: I don't think there's a rule or something. I'm also fine with multiple exit points. Maybe here's an article which prefers single returns: https://stschindler.medium.com/having-only-one-return-statement-is-better-and-more-safe-5cd379b7e87c

michaelworm: Unfortunately it's behind a "registration wall" :(

stephanschroeter (approved):

PR #3327, opened 2024-06-05
[BUG] - Accessible name is missing if Segmented Control is used with icon only (Issue/3190)

### LINKED ISSUE #3190: [BUG] - Accessible name is missing if `Segmented Control` is used with icon only
### JS Framework

all

3.15.0-rc.1

all

all

### Browser version

all

### Reproduction Example

Stackblitz: https://stackblitz.com/edit/ohvkiq?file=index.html

1. Go to https://stackblitz.com/edit/ohvkiq?file=index.html
2. Open Dev Console
3. Highlight the `<button>` element
4. Look into accessibility tree
5. See error: the accessible name is empty

### Expected behavior

If icons only are used in `Segmented Control` the user should have the ability to add an corresponding `aria` property to provide an accessible name.

**Hint:** Also re-activate axe-core test in `segmented-control.html` vrt test page!

o-hook:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3190/components/segmented-control/examples#with-icons

#### Scope

#### Resolved Issue

Resolves #3190


packages/components/src/components/segmented-control/segmented-control-item/segmented-control-item.tsx
```diff
@@ -82,7 +96,7 @@ export class SegmentedControlItem {
     const PrefixedTagNames = getPrefixedTagNames(this.host);
 
     return (
-      <button type="button" {...getButtonAttributes(this.host.selected, this.disabled)}>
+      <button type="button" {...getSegmentedControlItemAriaAttributes(this.host.selected, this.disabled, this.aria)}>
```
henrijoss, line 99: Should we validate if an `aria-label` is set when no content is passed?

o-hook: @henrijoss good approach, but we should think about validating against accessible names in a more generic way instead of adding custom validations for every use case (also this validation is missing on some other components, like link, button, etc.)

Google provides some (unofficial) package for calculating an HTML element's accessible name: https://github.com/google/accname 

But this would be a separate task. What do you think?

henrijoss: Don't know about the package but I also think a more generic way of making sure the accessible names are provided would be good. Do we already have a ticket for this?

o-hook: Nope, but I will add one :-)

Issue #3338, opened 2024-06-11
[BUG]: Contents of `Toast` component are not announced in screen reader


o-hook:
Scope
---
Though the `Toast` component was developed with accessibility best practices in mind (including the usage of `role="status"`, the contents of the rendered `Toast` component are not announced in screen reader. This is due to the insufficient implementation of `role="status"` attribute.

Out of Scope
---

Notes  
---
- ARIA live attributes should always be rendered initially to have the best support for assistive technologies.
- Possible solution: try to add `role="status"` to the `host` element or with `ElementInternals` API

Acceptance criteria
---
- [ ] Contents of `Toast` component get announced in screen reader when host is rendered initially.

Subtasks
---
- [ ] Task

o-hook:
Will be fixed with #3130

PR #3356, opened 2024-06-25
Migrate Toast to be rendered on #top-layer (Issue/3130)

### LINKED ISSUE #3130: [FEATURE] Migrate Toast to be rendered on `#top-layer`
Scope
---
We've refactored the Flyout, Modal and Banner component to make use of `#top-layer` which causes the Toast not to be positioned on top of the mentioned components. Therefor, we need to ensure that the Toast is getting rendered on `#top-layer` too.

Notes  
---
- Ensure Toast is rendered on `#top-layer` (which is not the case when `<dialog />` gets opened by `.show()`)
- Consider using attribute `popover` on either a `<div />` or custom element `:host` or `<dialog />` opened by `.show()` in combination with attr `popover`… or maybe CSS `overlay: auto` is good enough
- https://developer.mozilla.org/en-US/docs/Web/CSS/overlay
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog?retiredLocale=de
- https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover
- https://blog.bitsrc.io/top-layer-vs-z-index-for-frontend-dialogs-fcb12217a834
- Take a look how Banner component was refactored to make use of `#top-layer` since they are kind of similar components
- Challenge the issue that "a popover on top of a modal dialog is currently not interactable":
  - https://github.com/whatwg/html/issues/9936
  - https://porschedev.slack.com/archives/C028Q9ALX7H/p1716384844395929?thread_ts=1716382640.357049&cid=C028Q9ALX7H
  - https://designsystem.porsche.com/latest/patterns/notifications/introduction > "Start Live Demo"

Acceptance criteria
---
- [x] Toast is rendered on `#top-layer`
- [x] [Patterns > Notifications @ Storefront](https://designsystem.porsche.com/v3/patterns/notifications/introduction) is updated accordingly (especially z-index table)
- [x] Also #3338 must be fixed

Subtasks
---
- [ ] Task

o-hook:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3130/

#### Scope

#### Resolved Issue

Resolves #3130 and #3338


packages/components/src/components/toast/toast/toast.tsx
```diff
@@ -55,7 +55,7 @@ export class Toast {
     const toast = toastManager.getToast();
 
     return (
-      <Host>
+      <Host role="status" aria-live="polite">
```
stephanschroeter, line 58: is it correct, that aria-live is always defined?

packages/components/src/components/toast/toast-item/toast-item.tsx
```diff
@@ -56,15 +63,15 @@ export class ToastItem {
...
+      <Host popover="manual">
         <PrefixedTagNames.pIcon
           class="icon"
           name={getInlineNotificationIconName(this.state)}
           color={`notification-${this.state}` as IconColor}
           theme={this.theme}
           aria-hidden="true"
         />
-        <p id={toastId} class="content" role="status" aria-live="polite" innerHTML={this.text}></p>
+        <p id={toastId} class="content" innerHTML={this.text}></p>
```
stephanschroeter, line 74: Do we still need class="content"? It might be also styled directly on p.

packages/components/src/components/toast/toast-item/toast-item.tsx
```diff
@@ -45,6 +46,12 @@ export class ToastItem {
     throwIfRootNodeIsNotOneOfKind(this.host, ['p-toast']);
   }
 
+  public componentDidRender(): void {
+    if (getHasNativePopoverSupport()) {
+      this.host.showPopover();
```
stephanschroeter, line 51: How does the focus behave in case a flyout with `<dialog/>` is opened which then triggers a toast?

packages/components/src/components/toast/toast-item/toast-item-styles.ts
```diff
@@ -7,15 +7,30 @@ import {
...
+import { addImportantToEachRule, getThemedColors, prefersColorSchemeDarkMediaQuery } from '../../../styles';
 
 export const getComponentCss = (state: ToastState, theme: Theme): string => {
   return getCss({
     '@global': {
       ':host': {
         display: 'grid',
+        opacity: 0, // needed to prevent flickering on initial render
+        maxWidth: 'inherit',
+        boxSizing: 'border-box',
```
stephanschroeter, line 19: why are some styles not wrapped into important helper?

packages/components/src/components/toast/toast/toast.tsx
o-hook: Needs to be on pre-rendered host to get announced in screen reader.

packages/components/src/components/toast/toast-item/toast-item.tsx
o-hook: As discussed: if the Toast is initialized inside of the `dialog` it is interactable. If it is initialized outside of `dialog` (e.g. end of `body`) then no interaction can be done. See also conversation on https://github.com/whatwg/html/issues/9936

packages/components/src/components/toast/toast-item/toast-item-styles.ts
o-hook: Adapted. But `opacity` needs to be overwritable due to the animation of the parent Toast element.

packages/components/src/components/toast/toast-item/toast-item.tsx
o-hook: fixed

stephanschroeter (approved):

Issue #3366, opened 2024-07-01
Extend Component Meta


henrijoss:
Scope
---
In preparation of a component configurator on the storefront, we need to add additional information to `componentMeta`.

Add SlotMeta
- `description`
- Prop relationship e.g. `heading`

PropMeta:
- `isControlled` indicates this prop value likely will be managed in an external state variable and changed dynmically from outside
- `mutable` indicates this prop is mutated from within the component

Improved SlotMeta:
```
type SlotMeta = {
  name: string; // Name of the slot. Empty name corresponds to the default slot.
  description: string;
  isRequired?: boolean; // Specifies if the slot is required. If undefined the slot is not required.
  allowedTagNames?: (TagName & HTMLElement)[]; // Specifies which tagNames are allowed to be used. If undefined all tags are allowed.
  hasAltProp?: boolean; // Specifies if the slot has an equal name prop which can be used instead.
};
```

ControlledStateMeta:
```
/**
 * Describes a property whose state is likely to be externally controlled and updated via a specific event.
 */
type ControlledStateMeta = {
  prop: string;
  event: string;
};
```

Related issues:
[#3315](https://github.com/porsche-design-system/porsche-design-system/issues/3315)

Out of Scope
---
List topics that are not part of this task.

Notes  
---
- Can we use Stencil Docs generation instead of our own logic? https://stenciljs.com/docs/docs-custom

Acceptance criteria
---
- [ ] Criteria

Subtasks
---
- [ ] Add `@slot` documentation
- [ ] Add `@controlled` documentation
- [ ] Add `@prop` documentation (CSS Custom Properties)
- [ ] Replace `component-meta` with https://stenciljs.com/docs/docs-custom
- [ ] Improve slot meta data structure 
- [ ] Add controlled state meta
- [ ] Add CSS Custom Properties
- [ ] Adjust usages
- [ ] Test to assure slot and controlled docs are in sync with component code?

PR #3368, opened 2024-07-01
Extend Component Meta (issue/3366)

henrijoss:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3366

#### Scope

#### Resolved Issue

Resolves [#3366](https://github.com/porsche-design-system/porsche-design-system/issues/3366)


packages/component-meta/scripts/generateComponentMeta.ts
```diff
@@ -272,82 +116,62 @@ const generateComponentMeta = (): void => {
...
+      let cleanedValue: boolean | number | string | object =
+        propValue === 'true'
+          ? true
+          : propValue === 'false'
+            ? false
+            : // undefined values get lost in JSON.stringify, but null is allowed
+              propValue
+                ?.replace(/^['"](.*)['"]$/, '$1') // propValue is a string and might contain a string wrapped in quotes since it is extracted like this
+                .replace(/\s+/g, ' ') // remove new lines and multiple spaces
+                .replace(/,( })/, '$1') ?? null; // remove trailing comma in original multiline objects
```
michaelworm, line 134: I know it was like this before but this is unreadable & unmaintainable. You should consider refactoring it.

packages/components/src/test-utils/addParentAndSetRequiredProps.ts
```diff
@@ -22,27 +21,35 @@ export const addParentAndSetRequiredProps = (tagName: TagName, component: any):
     component[childTagName] = child;
   }
 
-  if (requiredNamedSlots) {
-    requiredNamedSlots.forEach(({ slotName, tagName: requiredTagName }) => {
-      const child = document.createElement(requiredTagName);
+  const requiredNamedSlots = Object.entries(slotsMeta ?? {}).filter(([, value]) => value.isRequired);
```
michaelworm, line 24: A little bit easier to read and to avoid calling filter on an empty array would be would be 
```js
const requiredNamedSlots = slotsMeta ?? Object.entries(slotsMeta).filter(([, value]) => value.isRequired);
```

and below 

```js
if (requiredNamedSlots && requiredNamedSlots.length > 0) {
```

packages/components/src/test-utils/addParentAndSetRequiredProps.ts
```diff
@@ -22,27 +21,35 @@ export const addParentAndSetRequiredProps = (tagName: TagName, component: any):
...
 
       component.host.appendChild(child);
     });
   }
 
+  const requiredProps = propsMeta
+    ? Object.entries(propsMeta)
+        .filter(([, value]) => value.isRequired)
+        .map(([key]) => key)
+    : undefined;
```
michaelworm, line 43: `propsMeta` is fallbacked by `{}` in line 5, when no value is passed, the filter will be run on an empty Object. This could be avoided. Also ternaries add complexity, better write it out for more readability.

packages/storefront/src/components/ComponentMetaOverview.vue
```diff
@@ -45,7 +45,94 @@
     }
 
     get body(): string {
```
michaelworm, line 47: The whole function is way too complex. Consider creating helper functions for the constants which require complex conditions.
Consider these best practices for your refactor: 
- Try to avoid ternaries as they are harder to read/understand. 
- Try to avoid such cases `Object.entries(propsMeta ?? {}).map` since the map function will be called at least once when `propsMeta` is `undefined` or `null` which is useless.
- `if/elseif/elseif/else` could be avoided by using returns in if-conditions inside of a function
- Move typings out of the function to the top of the file eg. `<
        T extends { isDeprecated?: boolean; isBreakpointCustomizable?: boolean; isExperimental?: boolean },
      >(`

Remember that other colleagues have to understand it so it's your challenge to support them by writing clean and structured code which is easily understandable. You already did a great job reducing the complexity after rewriting but there is more potential.

michaelworm (changes requested):

packages/storefront/src/components/ComponentMetaOverview.vue
```diff
@@ -45,7 +45,94 @@
...
+        if (Array.isArray(value)) {
+          return value
+            .sort()
+            .map((val) => (rowKey === 'nestedComponents' ? val : `<code>${val}</code>`))
+            .join('<br>');
+        }
+        return value;
+      };
+
+      const rowKeys = [
```
michaelworm, line 135: The whole function is way too complex. Consider creating helper functions for the constants which require complex conditions.
Consider these best practices for your refactor: 
- Try to avoid ternaries as they are harder to read/understand. 
- Try to avoid such cases `Object.entries(propsMeta ?? {}).map` since the map function will be called at least once when `propsMeta` is `undefined` or `null` which is useless.
- `if/elseif/elseif/else` could be avoided by using returns in if-conditions inside of a function
- Move typings out of the function to the top of the file eg. `<
        T extends { isDeprecated?: boolean; isBreakpointCustomizable?: boolean; isExperimental?: boolean },
      >(`

Remember that other colleagues have to understand it so it's your challenge to support them by writing clean and structured code which is easily understandable. You already did a great job reducing the complexity after rewriting but there is more potential.

michaelworm (changes requested):

stephanschroeter (approved):

PR #3443, opened 2024-08-08
Issue/3385

### LINKED ISSUE #3385: Create `p-textarea`
Scope
---
Currently, we're providing `p-textarea-wrapper` which requires a `textarea` as child element. To provide a more solid/modern variant we should create a new component called `p-textarea` with a `textarea` within the Shadow DOM and the same feature scope like `p-textarea-wrapper`.

Notes  
---
- See **Select** and **Multi Select** for reference
- Check if `formAssociated / attachInternals` is already usable https://stenciljs.com/docs/form-associated 
- Can be implemented as either a controlled or an uncontrolled component.
- Ensure native attributes like e.g. `required` don't interfere with custom prop `required`, find a pattern how to deal with it.
- Check how to deal with line breaks? Or other HTML entities within native textarea text node?
- Besides interface proposal, check if other native `textarea` attributes should be exposed as `p-textarea` component prop, https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea
- Interface proposal:
```
# Name: 
p-textarea

# Props:
name: string;
value?: string = ''; // mutable? or as alternative use a text node like it's the case natively
placeholder: string = '';
disabled: boolean = false;
required: boolean = false;
readonly: boolean = false; // or readOnly?
dirname: string = undefined; // or dirName?
autofocus: boolean = false; // or autoFocus?
form: string;
maxlength: number; // or maxLength?
minlength: number; // or minLength?
autocomplete: 'on' | 'off' | string // or autoComplete?
spellcheck: 'true' | 'default' | 'false'; // or spellCheck?
wrap: 'hard' | 'soft' | 'off'

loading: boolean = false; // experimental prop, check if real use cases exist before implementing
label: string = '';
hideLabel: boolean | BreakpointCustomizable<boolean> = false;
description: string = '';
message: string = '';
showCounter: boolean = false;
state: 'none' | 'error' | 'success' = 'none';
theme: Theme = 'light';

# Events:
update(CustomEvent< {name: string; value: string;}>)

# Slots:
- slot="label"
- slot="description"
- slot="message"
```

Acceptance criteria
---
- [ ] `p-textarea` is provided by npm package `@porsche-design-system/components-{js|angular|react|vue}` with the same feature scope like `p-textarea-wrapper` and supports the most important native `textarea` attributes
- [ ] `p-textarea` is documented at https://designsystem.porsche.com
- [ ] ~~`p-textarea-wrapper` is deprecated~~

Subtasks
---
- [ ] Task

Questions
---
- Currently `onInput` used for updating of textarea value, causing lots of lifecycles
- A11y
- React event typing

Decisions
---
- No slotted content allowed, instead we use a value attribute and pass it through to native textarea. Biggest difference to native textarea.
- `dirname` not working currently, maybe stencil `formAssociated` does not support it -> remove it
- Native html validation not supported? -> Not yet

ElementInternals API
---
https://stenciljs.com/docs/form-associated
https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals
https://developer.mozilla.org/en-US/docs/Web/HTML/Constraint_validation

`@AttachInternals() private internals: ElementInternals;`
Stencil Annotation automatically calls `attachInternals` and provides `ElementInternals` interface to web component privately.  Nothing is automatically exposed. 

FormData:
`value` of `FormData` object can be set with `internals.setFormValue()`. As long as a `name` attribute is defined on the web component, the value will now be included in the formData submission.

Reset:
This function will be triggered when the form is reset (Either via reset button or form.reset()):
```
public formResetCallback(): void {
    this.internals.setValidity({});
    this.internals.setFormValue('');
  }
```

Validation:
`checkValidity()` and `reportValidity()` can be exposed publically by Stencil with `@Method` decorator but won't help in the native Form Validation API. The `validity` getter can't be exposed in stencil currently but you can set a Validation State manually by using the `this.internals.setFormValidity()`. 

E.g. for `required` prop. Last param is the anchorElement which will receive focus when the form is submitted and validation fails.
```
this.internals.setValidity({
        valueMissing: true
    }, "Custom message", 
this.textAreaElement);
```

kaiszybiak:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3385/

#### Scope

Introduces Textarea Component

#### Resolved Issue

Resolves #3385


packages/components/src/components/textarea/textarea.examples.md
```diff
@@ -0,0 +1,183 @@
...
+  get stateMarkup() {
+    const attr = `message="${this.state !== 'none' ? `Some ${this.state} validation message.` : ''}"`;
+    return `<p-textarea aria-invalid="${this.state === 'error'}" name="some-name" label="Some label" value="Some value" state="${this.state}" ${attr}></p-textarea>`;
+    }
+    
+  slots = `<p-textarea name="some-name" aria-labelledby="some-label-id" aria-describedby="some-description-id some-message-id" state="error">
+  <span slot="label" id="some-label-id">Some label with a <a href="https://designsystem.porsche.com">link</a>.</span>
+  <span slot="description" id="some-description-id">Some description with a <a href="https://designsystem.porsche.com">link</a>.</span>
+  <span slot="message" id="some-message-id">Some error message with a <a href="https://designsystem.porsche.com">link</a>.</span>
+</p-textarea>`;
```
o-hook, line 165: Because we handle all the ARIA relationship inside of the component, we don't need to set all those extra ARIA attributes on the host element anymore. And in consequence, you can also remove the ID attributes from the Slots  as well.

packages/components/src/components/textarea/textarea.examples.md
```diff
@@ -0,0 +1,183 @@
+<ComponentHeading name="Textarea"></ComponentHeading>
+
+The `p-textarea` component is a multi-line text input control. Don't use a Textarea component if you want to allow users
+to enter shorter responses that are no longer than a single line, such as a phone number or name. In this case, you
+should use the Text Field component.
+
+A `label` is a caption which informs the user what information a particular form field is asking for. The `p-textarea`
+component can be used with or without a label, but it's recommended to keep the label visible for better accessibility
+whenever possible. When used without a label, it's best practice to provide a descriptive label text for screen readers.
```
o-hook, line 9: "~~it's best practice~~" should be "it's mandatory"

packages/components/src/components/textarea/textarea.examples.md
```diff
@@ -0,0 +1,183 @@
...
+
+If using **slotted contents** to serve form elements, make sure to provide the right **ARIA attributes** to give screen
+reader users the corresponding information:
+
+1. Add a unique ID to the `slot="label"` element
+1. Add a unique ID to the `slot="message"` element (if they are created)
+1. Add corresponding `aria-labelledby="some-label-id"` to the `textarea` element which points to the `label` ID
+1. Add corresponding `aria-describedby="some-description-id some-message-id"` to the `textarea` element which points to
+   both, the `description` ID (if set) and the `message` ID when the (error/success) message appears
+
```
o-hook, line 107: All this A11y hints can also be removed.

kaiszybiak: removed

packages/components/src/components/textarea/textarea.examples.md
kaiszybiak: updated

packages/components/src/components/textarea/textarea.examples.md
kaiszybiak: updated

packages/components/src/components/textarea/textarea-utils.ts
```diff
@@ -0,0 +1,12 @@
+import type { FormState } from '../../utils/form/form-state';
+export type TextareaState = FormState;
+
+export const AUTO_COMPLETE = ['off', 'on', ''] as const;
+export type TextareaAutoComplete = (typeof AUTO_COMPLETE)[number];
+
+export const TEXTAREA_WRAPS = ['hard', 'soft', 'off'] as const;
+export type TextareaWrap = (typeof TEXTAREA_WRAPS)[number];
```
henrijoss, line 8: Why not use `lib.dom.d.ts` types here?

packages/components-js/tests/e2e/specs/textarea.e2e.ts
```diff
@@ -0,0 +1,441 @@
...
+    await expect(textarea).toHaveValue('');
+
+    await page.locator('button[type="submit"]').click(); // Check if ElementInternal value was reset as well
+
+    expect((await getEventSummary(form, 'submit')).counter).toBe(1);
+    expect(await getFormDataValue(form, name)).toBe('');
+  });
+});
+
+test.describe('focus state', () => {
```
henrijoss, line 261: Add tests for programmatic `focus()` of host and `autoFocus` prop

packages/components-js/tests/e2e/specs/default-dom.e2e.ts-snapshots/p-textarea.txt
```diff
@@ -0,0 +1,13 @@
+<div class="root">
+  <label class="label" id="label" for="textarea"> </label>
```
stephanschroeter, line 2: Why do we need id="label"?

packages/components/src/components/textarea/textarea-styles.ts
```diff
@@ -0,0 +1,146 @@
...
+        ['textarea:not(:disabled):not(:focus):not([readonly]):hover,label:hover~.wrapper textarea:not(:disabled):not(:focus):not([readonly])']:
+          {
+            borderColor: formStateHoverColor || primaryColor,
+            ...prefersColorSchemeDarkMediaQuery(theme, {
+              borderColor: formStateHoverColorDark || primaryColorDark,
+            }),
+          },
+      }) as Styles),
+      // TODO: getFocusJssStyle() can't be re-used because focus style differs for form elements
+      ['textarea:focus']: {
```
stephanschroeter, line 98: I think `[…]` is not needed? It's also possible to work with `&:focus` instead. Applies also to other selectors in this file.

packages/components-js/tests/e2e/specs/default-dom.e2e.ts-snapshots/p-textarea.txt
henrijoss: image

For the htmlFor association

henrijoss: Defined in our Label FC. Unused in this case but needed for aria-labelledby in other components

stephanschroeter (approved):

PR #3460, opened 2024-08-15
feat(components): make readyState of componentsReady configurable

### LINKED ISSUE #3394: Improve `componentsReady`
Scope
---
Teams aiming for maximum performance of their web application are looking for a way to detect as early as possible when PDS components have been bootstrapped. Therefor, we should evaluate if it's necessary to wait for `complete` readiness of document within `componentsReady` (`packages/components/src/components-ready.ts`) instead of e.g. `loading`. Alternatively, we could even make the `document.readyState` check configurable.

Notes  
---
Screenshot 2024-07-10 at 14 57 16

Acceptance criteria
---
- [ ] `componentsReady` is resolving its promise as early as possible as soon as PDS components have been bootstrapped.

Subtasks
---
- [ ] Task

kaiszybiak:
…ks | #3394

### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3394/

#### Scope

#### Resolved Issue

Resolves #3394


packages/components/src/components-ready.ts
```diff
@@ -3,21 +3,30 @@ import type { PorscheDesignSystem } from './types';
...
+const documentReadyStateHierarchy: Record<DocumentReadyState, number> = {
+  loading: 0,
+  interactive: 1,
+  complete: 2,
+};
+
+export const componentsReady = (
+  el: HTMLElement = document.body,
+  readyState: DocumentReadyState = 'complete'
+): Promise<number> => {
```
leifmarcus, line 15: This might lead to problems, because `componentsReady` is often used without the first parameter. Adding another parameter might break some implementations.

A solution would be to make the first parameter of type `HTMLElement | DocumentReadyState`. Then you could check, if it is an element or if it is the ready state that is passed. If you would only want to set the ready state, but not the html element, you would need to call the function like `componentsReady(undefined, 'complete')` which is not ideal I guess.

Also, if you would pass `null`, the default value would not work anymore.
Something like this might work. In this case, the function can be used only with the ready state, with ready state and element or with element only.

```suggestion
export const componentsReady = (
  el: HTMLElement | DocumentReadyState = document.body,
  readyState: DocumentReadyState = 'complete'
): Promise<number> => {
  const targetNode = el && 'tagName' in el ? el : document.body;
  const targetState = readyState ? readyState : 'complete';
```

henrijoss: Adding another parameter shouldn't be able to break any implementation in this case. Passing undefined as first parameter should be fine. Also you can just call it with the document.body if you want it to be more clear `componentsReady(document.body, 'complete')`.

Changing the parameter intention depending on the passed types makes the function very unintuitive. The function parameter names don't make sense anymore and it adds unnecessary complexity which might lead to bugs/problems. This would be valid as well `componentsReady('loading', 'complete')`.

stephanschroeter (approved):

packages/components/src/components-ready.ts
leifmarcus: This was just more about Developer experience. If you only want to pass the ready state, you have to always pass `undefined` even if a not available element in Frontend Frameworks will be mostly defined as `null`. 

BTW: If you use typescript with overload, then `componentsReady('loading', 'complete')` is not possible anymore.

leifmarcus:
```typescript
// using overloads to disallow the false usage:

interface IOverloadFn {
  (state?: DocumentReadyState): Promise<void>;
  (el?: HTMLElement, state?: DocumentReadyState): Promise<void>;
}

const node = document.createElement('div');

export const componentsReady: IOverloadFn = async (
  elm?: HTMLElement | DocumentReadyState,
  state?: DocumentReadyState,
): Promise<void> => {
  const target = elm && 'tagName' in (elm as HTMLElement) ? (elm as HTMLElement) : document.body;
  const targetState = typeof elm === 'string' ? elm : (state ?? 'complete');
};

componentsReady('complete', 'loading'); // is not allowed
componentsReady(node, 'loading');       // is allowed
componentsReady('loading');             // is allowed
componentsReady();                      // is allowed
componentsReady(node);                  // is allowed
```

kaiszybiak: Hi @leifmarcus 🙂,

Thank you for your suggestion and for thinking about ways to enhance our approach. We appreciate the thoughtfulness behind your proposal. After some discussion, we’ve decided to continue with the current method. While both approaches have their merits, we believe that keeping the function signature clear and requiring explicit undefined values, though slightly less convenient, better supports clarity and maintainability.

PR #3472, opened 2024-08-21
Further improve UI of Form Elements

stephanschroeter:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/housekeeping/improve-ui-form-elements

#### Scope

#### Resolved Issue

Resolves [IssueNumber]


packages/components/src/components/textarea/textarea-styles.ts
```diff
@@ -71,6 +71,8 @@ export const getComponentCss = (
         background: 'transparent',
         textIndent: 0,
         color: primaryColor,
+        // min width is needed for showing at least 1 character in very narrow containers. The "1rem" value is the minimum safe zone to show at least 1 character.
+        minWidth: `calc(1rem + ${formElementPaddingHorizontal}*2 + ${borderWidthBase}*2)`,
```
henrijoss, line 75: What about using `1ch`?

o-hook: @henrijoss:  Just using `1ch` doesn't solve the problem in general, because `1ch` reflects just the width of the `0` character and this doesn't fit the `M` character for example. But yes, we have to investigate further what's better: `ch` or `rch` or `rem` as unit.

henrijoss (dismissed):

o-hook (approved):

PR #3510, opened 2024-09-25
[BUG] - Modal doesn't show scrollbar under certain pre conditions

### LINKED ISSUE #3509: [BUG] - Modal doesn't show scrollbar under certain pre conditions
### Contact Details

_No response_

### JS Framework

All

3.18.0

MacBook Pro

Firefox, Chrome

### Browser version

_No response_

### Reproduction Example

https://porschedev.slack.com/archives/CT7AVHTTQ/p1726656935319749

1. Go to OS settings and activate "show scrollbars always"
2. Go to https://moduleguide.pcom.weu.porsche.com/__sandbox.html?storyId=src-modules-features-features-story-vue&variantId=src-modules-features-features-story-vue-0
3. Click on 'open drawer'
4. See error, Modal doesn't show any scrollbar in Chrome or Firefox while Safari works

Screenshot 2024-09-25 at 10 54 02

### Expected behavior

Modal shall show scrollbar correctly in Chrome and Firefox

### Relevant log output

_No response_

kaiszybiak:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3509/

#### Scope

#### Resolved Issue

Resolves #3509


packages/components/src/utils/setScrollLock.ts
henrijoss, line 1: Since the styles aren't applied directly on the element but with this css file, any inline styles on the `html`, `body` will take precedence over this?

Since this is a non standard css selector, how does this behave on mobile? Please also check in BrowserStack

kaiszybiak: > Since the styles aren't applied directly on the element but with this css head style tag, any inline styles on the html, body will take precedence over this?

We could add `!important` to each style to enforce it.

```SCSS
body::-webkit-scrollbar {
  display: none; /* hide scrollbar in WebKit browsers */
}
```

Is the recommended way to hide Scrollbars: https://www.w3schools.com/howto/howto_css_hide_scrollbars.asp for Chrome, Safari and Opera however only using `scrollbar-width: none;` should also work. 

So maybe something like this would be enough:

```tsx
export const setScrollLock = (isOpen: boolean): void => {
  document.body.style.overflow = isOpen ? 'hidden' : '';
  document.documentElement.style.overflow = isOpen ? 'hidden' : '';
  document.body.style.scrollbarWidth = isOpen ? 'none' : '';
  document.documentElement.style.scrollbarWidth = isOpen ? 'none' : '';
};
```

kaiszybiak:
Closed because https://issues.chromium.org/issues/40835364

PR #3542, opened 2024-10-16
Form reset + ElementInternals API for Form Components [Select/MultiSelect]

### LINKED ISSUE #3455: Form reset + ElementInternals API for Form Components
Scope
---
In the `p-textarea` component we added support for the `form.reset()` and `<button type="reset">` functionality when used within a form context. We should make sure this feature also works for our other custom form components (`p-select`, `p-multi-select`, `p-pin-code`).

Notes  
---
- `p-textarea` uses the `ElementInternals API` while the `p-select`, `p-multi-select`, `p-pin-code` use a hidden select in the light dom to make the native form submission work.
- See `p-textarea` issue [#3385](https://github.com/porsche-design-system/porsche-design-system/issues/3385)
- To not be breaking we need to ensure this issue gets resolved too: https://github.com/porsche-design-system/porsche-design-system/issues/3508
- If it makes sense split PR into multiple ones per refactored component

Acceptance criteria
---
- [ ] `form.reset()` and `<button type="reset">` work when using the `p-select` or `p-multi-select` in a form context.
- [ ] `p-select`, `p-pin-code` and `p-multi-select` rely on `ElementInternals API` instead of custom synced and hidden select element

Subtasks
---
- [ ] Task

kaiszybiak:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3455_select/

#### Scope

- Use ElementInternals API for select and remove native select component.

#### Resolved Issue

Resolves #3455

#### Notes

- ~~Do we want to expose the `selected` attribute for the `p-select-option` similar to native select `option`?~~
- ~~If no value is provided the native select `option` will take the `innerHTML` as value. Do we want to mimic this behavior?~~
- ~~With the current implementation we have no access to the `validityState` as we removed the native `select`. Therefore e.g. you can still submit the form if the select value is undefined but required unlike for native select.~~ 
- ~~In the `select-utils.ts` we have `updateSelectOptions()`. There we do:~~

  ```ts
  // Option without value for empty selection
  const optionToSelect = options.find((option) => option.value === undefined);
  ```
  ~~If we want to mimic native select behavior it should be:~~
  
  ```ts
  
  export const determineSelectedOption = (options: SelectOption[]): SelectOption | undefined =>
  options.find((option) => option.value === undefined && !option.disabled) ||
  options.find((option) => !option.disabled);
  
  // Selects the first option with an undefined value; otherwise, defaults to the first option.
  const optionToSelect = determineSelectedOption(options);
  ```


packages/components/src/components/multi-select/multi-select/multi-select.tsx
```diff
@@ -130,21 +128,24 @@ export class MultiSelect {
   /** Adapts the select color depending on the theme. */
   @Prop() public theme?: Theme = 'light';
 
+  /** The id of a form element the select should be associated with. */
```
henrijoss, line 131:
```suggestion
  /** The id of a form element the multi-select should be associated with. */
```

packages/components-js/tests/e2e/specs/multi-select.e2e.ts
```diff
@@ -135,7 +132,7 @@ const initMultiSelect = (page: Page, opt?: InitOptions): Promise<void> => {
   const selectOptions = [...'abc', ...(amount === 5 ? 'de' : '')]
     .map((x, idx) => {
       const attrs = [disabledIndex === idx ? 'disabled' : ''].join(' ');
-      const option = `<p-multi-select-option value="${x}" ${attrs}>Option ${x.toUpperCase()}</p-multi-select-option>`;
+      const option = `<p-multi-select-option ${x ? `value="${x}"` : ''} ${attrs}>Option ${x.toUpperCase()}</p-multi-select-option>`;
```
henrijoss, line 135: How can `x` be falsy here?

packages/components-js/tests/e2e/specs/multi-select.e2e.ts
```diff
@@ -1480,3 +1231,118 @@ test.describe('optgroups', () => {
...
+    const form = getForm(page);
+    await setProperty(multiSelect, 'value', value);
+
+    await addEventListener(form, 'submit');
+    expect((await getEventSummary(form, 'submit')).counter).toBe(0);
+
+    await page.locator('button[type="submit"]').click();
+
+    const test = await getEventSummary(form, 'submit');
+    console.log('test', test.targets);
```
henrijoss, line 1253:
```suggestion
```

henrijoss, line 1253:
```suggestion
```

packages/components/CHANGELOG.md
```diff
@@ -19,6 +19,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0),
 - `Checkbox`, `Textarea`: `formDisabledCallback` and `formStateRestoreCallback` from ElementInternals API and sync
   validity with form element. ([#3528](https://github.com/porsche-design-system/porsche-design-system/pull/3528))
 
+#### Changed
```
henrijoss, line 22: - Add to `Added` section that the form prop to associate the elements with a form was added
- Add Note that native validation won't work anymore

packages/components-js/tests/e2e/specs/select.e2e.ts
```diff
@@ -1633,3 +1356,116 @@ test.describe('optgroups', () => {
     }
   });
 });
+
+test.describe('form', () => {
+  test('should include name & value in FormData submit', async ({ page }) => {
```
henrijoss, line 1361: We should also test if this works when the value is changed dynamically either by keyboard, mouse or programmatically. Same for `multi-select`

packages/components/src/components/multi-select/multi-select/multi-select.tsx
```diff
@@ -166,40 +167,50 @@ export class MultiSelect {
...
-        updateNativeOptions(this.nativeSelect, this.multiSelectOptions);
+    }
+  }
+
+  @Watch('form')
+  public updateFormAssociation(): void {
+    if (this.form) {
+      const formElement = document.getElementById(this.form) as HTMLFormElement;
+      if (formElement) {
+        formElement.appendChild(this.host);
```
henrijoss, line 185: Is this correct? Doesn't seem like a good idea to change the DOM here. Isn't there a way to associate with a new form?

packages/components/src/components/multi-select/multi-select/multi-select.tsx
```diff
@@ -166,40 +167,50 @@ export class MultiSelect {
...
     // Use initial value to set options
     setSelectedOptions(this.multiSelectOptions, this.value);
-    if (this.isWithinForm) {
-      this.nativeSelect = initNativeMultiSelect(this.host, this.name, this.disabled, this.required);
-      updateNativeOptions(this.nativeSelect, this.multiSelectOptions);
-    }
   }
 
   public componentDidLoad(): void {
+    this.setFormValue(this.value);
```
henrijoss, line 213: Is it necessary to set this here? Why not in connected or willLoad

packages/components-js/tests/e2e/specs/multi-select.e2e.ts
kaiszybiak: I thought about making the `selectOptions` configurable. With the current setup it can't be falsy, so I reverted the change

packages/components/src/components/multi-select/multi-select/multi-select.tsx
kaiszybiak: After testing it doesn't seem to be necessary. Not sure under which circumstances I added it 🤔. I tested it outside the form and it works just fine, also the e2e test succeeds, so I removed it.  Same for select

packages/components/src/components/multi-select/multi-select/multi-select.tsx
kaiszybiak: updated

packages/components-js/tests/e2e/specs/multi-select.e2e.ts
```diff
@@ -1480,3 +1231,118 @@ test.describe('optgroups', () => {
...
+    await addEventListener(form, 'submit');
+    expect((await getEventSummary(form, 'submit')).counter).toBe(0);
+
+    await page.locator('button[type="submit"]').click();
+
+    expect((await getEventSummary(form, 'submit')).counter).toBe(1);
+    expect(await getFormDataValues(form, name)).toStrictEqual(value);
+  });
+
+  test('should reset multi-select value to its initial value on form reset', async ({ page }) => {
```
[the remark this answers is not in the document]
kaiszybiak: added a react e2e test for both select and multi-select

packages/components/src/components/select/select/select.tsx
```diff
@@ -128,20 +126,23 @@ export class Select {
   /** Adapts the select color depending on the theme. */
   @Prop() public theme?: Theme = 'light';
 
+  /** The id of a form element the select should be associated with. */
+  @Prop() public form?: string;
```
henrijoss, line 130: How does the form association work now? Also an E2E test for his would be good (Same for `multi-select`)

henrijoss (approved):

PR #3576, opened 2024-10-29
Housekeeping/uxpin dummy img

robhil:
### Pull Request Checklist

- [X] I have read and accept the [Contributor License Agreement](https://opensource.porsche.com/docs/cla)

#### References

Sizing of DummyImage not working. And for UX Pin Image no URL instead of upload is working

#### Scope

Make component resizable when is added as root component in uxpin
Add objectFit property
Use image control which allows upload image

https://github.com/user-attachments/assets/4973598d-3908-436b-adf4-6a81a28b01a3

### DIFF
--- a/packages/components-react/projects/uxpin-wrapper/package.json
+++ b/packages/components-react/projects/uxpin-wrapper/package.json
@@ -22,7 +22,7 @@
     "@babel/preset-typescript": "^7.24.7",
     "@porsche-design-system/assets": "5.0.2",
     "@types/react": "18.3.3",
-    "@uxpin/merge-cli": "3.4.3",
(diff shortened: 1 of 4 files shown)


marcelbertram (changes requested):
Works as expected but with focus on Dummy elements, could you please hide the not needed ones: DummyButton, DummyCheckbox, DummySelect, Dummy Textarea as well?

I think for the moment we still need DummyTextField and DummyRadioButton for the Wrapper Components TextField and Radio Button. If you can hide them as well without breaking this components this would be also great.

marcelbertram (approved):

PR #3585, opened 2024-10-31
handle onClick property in crest component, bring back DummyButton as…

robhil:
… it is used in Tabs Bar

### Pull Request Checklist

#### References

- Crest/Link/Link Pure: No default href and click event if not defined. Should be set manually using props

- Tabsbar not working and breaking prototype

#### Scope

I added onClick property to crest. When onClick is defined I set href to '#' to make crest link

For link and link pure I did nothing  as click event is added and href is ignored.

Additionally, I hide `target` property as this does not work when we use onClick.

For TabsBar I had to bring back DummyButton as this component used this in preset so I can't hide it.

https://github.com/user-attachments/assets/079900a8-910f-4ff3-aa5b-69b5226796d7


marcelbertram (changes requested):
Link Pure should also have a pointer cursor if a click event is defined:
https://share.zight.com/5zuokdDD

marcelbertram (approved):

PR #3592, opened 2024-11-05
[BUG] Carousel pagination breaks after change in slides-per-page (Issue/3551)

### LINKED ISSUE #3551: [BUG] Carousel pagination breaks after change in slides-per-page
When the `slidesPerPage` prop changes dynamically (like its the case when using it as a breakpoint customizable value), the pagination will break after this change. Only happens when `focusOnCenterSlide` prop is used. Likely because newly introduced `getPageCount` function isn't used in the `updateAmountOfPages` function.

### JS Framework

- All

- 3.20.0

All

All

### Reproduction Example

- https://stackblitz.com/edit/nxbess?file=index.html

1. Go to https://stackblitz.com/edit/nxbess?file=index.html
2. Resize viewport of example to show 2 slides per page
3. Navigate through slides
4. See broken pagination

### Expected behavior

- Pagination works after window resize

### Notes

Carousel has to be used with `focusOnCenterSlide` for the problem to occur. Introduced with https://github.com/porsche-design-system/porsche-design-system/issues/3438

Probably because `renderPagination` doesn't use `this.getPageCount()` here:

```
  private updateAmountOfPages = (): void => {
    this.amountOfPages = getAmountOfPages(
      this.slides.length,
      // round to sanitize floating numbers
      this.slidesPerPage === 'auto' ? 1 : Math.round(getCurrentMatchingBreakpointValue(this.slidesPerPage))
    );
    renderPagination(this.paginationEl, this.amountOfPages, this.splide?.index || 0, this.splide);
  };
```

o-hook:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3551

#### Scope

#### Resolved Issue

Resolves #3551


packages/components-js/tests/e2e/specs/carousel.e2e.ts
```diff
@@ -610,6 +634,29 @@ test.describe('viewport change', () => {
     await waitForStencilLifecycle(page);
     expect(await pagination.evaluate((el) => el.children.length)).toBe(4);
   });
+
+  test('should update pagination for BreakpointCustomizable in focusOnCenterSlide mode', async ({ page }) => {
```
henrijoss, line 638: The problem here wasn't the breakpoint customisable but rather the dynamic change to the `slidesPerPage`. Do would be enough to just change this value dynamically.

packages/components/CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0),
 
 #### Fixed
 
+- `Carousel`: rendering of pagination in case of `focusOnCenterSlide` prop is set
```
henrijoss, line 24:
```suggestion
- `Carousel`: dynamic change in `slidesPerPages` when using `focusOnCenterSlide` prop breaks pagination
```

packages/components-js/tests/e2e/specs/carousel.e2e.ts
o-hook: Plz re-check :-)

packages/components-js/tests/e2e/specs/carousel.e2e.ts
```diff
@@ -229,6 +229,24 @@ test('should update pagination on next button clicks', async ({ page }) => {
   expect(await getCssClasses(bullet3)).toBe('bullet');
 });
 
+test('should update pagination in focusOnCenterSlide mode if slidesPerPage changed dynamically', async ({ page }) => {
```
henrijoss, line 232: Maybe would be good to check if the pagination also still works and not only if the amount if correct?

henrijoss (approved):

PR #3646, opened 2024-11-25
Housekeeping/carousel a11y improvements

o-hook:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/housekeeping/carousel-a11y-improvements/components/carousel/examples

#### Scope

Change active element on tabbing to internal slide element. This results in a better screen reader output with less verbosity.

#### Resolved Issue

Resolves housekeeping


packages/components/src/components/carousel/carousel.tsx
```diff
@@ -180,6 +180,7 @@ export class Carousel {
   private btnNext: HTMLPButtonPureElement;
   private paginationEl: HTMLElement;
   private slides: HTMLElement[] = [];
+  private splideSlides: HTMLElement[] = [];
```
henrijoss, line 183: You can also create a private getter instead of storing it in a private member
```suggestion
  private get splideSlides(): HTMLElement[] {
    return this.splide.Components.Elements.slides;
  }
```

henrijoss (approved):

PR #3648, opened 2024-11-26
[FEATURE] Provide the "form" attribute on p-button & p-button-pure

### LINKED ISSUE #3149: [FEATURE] Provide the "form" attribute on p-button
Scope
---
Sometimes placing a button outside of the form element might be useful, for this case there is the ["form" attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button?retiredLocale=de#form) for the native button element. PDS currently does not support using this property on p-button, instead the same behaviour can be mirrored by using a custom click event targeting the form and submitting it manually. We could introduce the "form" property to the p-button to enable this functionality natively. 

https://porschedev.slack.com/archives/CT7AVHTTQ/p1711349284843719

Notes  
---
Probably we have issues defining a form attribute on the button which points to an id of the form because it's enclosed within a shadowDom and references between light- amd shadowDom are not possible. `improveButtonHandlingForCustomElement()` might help to solve the situation.

Acceptance criteria
---
- [ ] The "form" attribute is enabled as a property on p-button
- [ ] Update Storefront with an example for Flyout, when the button is placed in the footer slot submitting the form inside the content slot

kaiszybiak:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3149/

#### Scope

related: https://github.com/porsche-design-system/porsche-design-system/issues/3657

#### Resolved Issue

Resolves #3149


packages/components-angular/src/app/examples/flyout-example-form.component.ts
```diff
@@ -0,0 +1,40 @@
...
+      <form id="some-form" (submit)="onSubmit($event)">
+        <p-checkbox name="some-checkbox" label="Some Label"></p-checkbox>
+        <p-textarea name="some-textarea" label="Some Label"></p-textarea>
+      </form>
+      <p-button-group slot="footer">
+        <p-button type="submit" form="some-form">Submit</p-button>
+        <p-button type="reset" [variant]="'secondary'" form="some-form">Reset</p-button>
+      </p-button-group>
+
+      <p-text slot="sub-footer" innerHTML="Last submitted data:<br/><br/> {{ lastSubmittedData }}"></p-text>
```
henrijoss, line 18:
```suggestion
      <p-text slot="sub-footer">
          Last submitted data:
          <br /><br />
          {{ lastSubmittedData }}
      </p-text>
```

packages/components-angular/src/app/examples/flyout-example-form.component.ts
```diff
@@ -0,0 +1,40 @@
...
+    this.isFlyoutOpen = true;
+  }
+  onDismiss() {
+    this.isFlyoutOpen = false;
+  }
+
+  onSubmit(e: Event): void {
+    e.preventDefault();
+    const formData = new FormData(e.target as HTMLFormElement);
+    this.lastSubmittedData = `checkbox: ${formData.get('some-checkbox') || 'none'} <br/>textarea: ${formData.get('some-textarea') || 'none'}`;
```
henrijoss, line 38: Would be better to just use two variables for checkboxValue and textAreaValue. Then you don't need to use innerHTML etc. Same for all other examples.

packages/components-react/tests/e2e/specs/flyout.e2e.ts
henrijoss, line 1: Since this is a feature of `p-button` and `p-button-pure` we should add some E2E in `button.e2e.ts` & `button-pure.e2e.ts`

packages/components/src/components/button-pure/button-pure.tsx
```diff
@@ -112,6 +114,11 @@ export class ButtonPure {
   /** Add ARIA attributes. */
   @Prop() public aria?: SelectedAriaAttributes<ButtonPureAriaAttribute>;
 
+  /** The id of a form element the button should be associated with. */
+  @Prop({ reflect: true }) public form?: string;
```
henrijoss, line 118: Add comment how reflect is necessary since it won't work in react without attribute

packages/components/src/components/button-pure/button-pure.tsx
```diff
@@ -123,6 +130,19 @@ export class ButtonPure {
...
+    if (this.form && this.internals.form) {
+      e.preventDefault();
+      switch (this.type) {
+        case 'submit':
+          this.internals.form.requestSubmit();
+          break;
+        case 'reset':
+          this.internals.form.reset();
+          break;
+      }
```
henrijoss, line 145: Would be better to just use if...else if here

packages/components/src/components/button-pure/button-pure.tsx
```diff
@@ -123,6 +130,19 @@ export class ButtonPure {
...
     if (this.isDisabledOrLoading) {
       e.stopPropagation();
+      return;
+    }
+
+    if (this.form && this.internals.form) {
+      e.preventDefault();
+      switch (this.type) {
+        case 'submit':
+          this.internals.form.requestSubmit();
```
henrijoss, line 140: Will this also send the button value? `requestSubmit` also takes in a `submitter` parameter where the button can be passed. It this necessary? Also add an E2E testing if the button value is included in the form.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit#submitter

packages/components/src/components/flyout/flyout.examples.md
henrijoss, line 1: The `button` and `button-pure` storefront pages should mention the use of ElementInternals and also provide a form example.

packages/components/src/components/button-pure/button-pure.tsx
henrijoss, line 1: How does `improveButtonHandlingForCustomElement` play together with `ElementInternals`? The custom solution should not be necessary anymore but maybe we want to keep it in case `ElementInternals` is not supported since otherwise submit won't work on older browsers?

packages/components/src/components/button/button.tsx
```diff
@@ -87,20 +89,42 @@ export class Button {
...
   private initialLoading: boolean = false;
 
   @Listen('click', { capture: true })
   public onClick(e: MouseEvent): void {
     if (isDisabledOrLoading(this.disabled, this.loading)) {
       e.stopPropagation();
+      return;
     }
+
+    if (this.form && this.internals.form) {
```
henrijoss, line 107: We also want to utilise `ElementInternals` when used inside of a form without the form attribute. Currently this is achieved by the `improveButtonHandlingForCustomElement ` function

packages/components/src/components/button/button.tsx
```diff
@@ -87,20 +89,42 @@ export class Button {
...
+      e.preventDefault();
+      if (this.type === 'submit') {
+        this.internals.form.requestSubmit();
```
henrijoss, line 110: Check if this really covers all needs: https://github.com/WICG/webcomponents/issues/814

kaiszybiak: https://github.com/porsche-design-system/porsche-design-system/issues/3657

packages/components/src/components/button-pure/button-pure.tsx
```diff
@@ -123,14 +131,30 @@ export class ButtonPure {
   public onClick(e: MouseEvent): void {
     if (this.isDisabledOrLoading) {
       e.stopPropagation();
+      return;
+    }
+
+    if (this.form && this.internals?.form) {
```
henrijoss, line 137: `improveButtonHandlingForCustomElement` is also executed when submitting

packages/components/src/components/button-pure/button-pure.tsx
```diff
@@ -123,14 +131,30 @@ export class ButtonPure {
...
+      e.preventDefault();
+      if (this.type === 'submit') {
+        this.internals?.form.requestSubmit();
```
henrijoss, line 140: Why optional chained?

henrijoss: `submitter` seems to be null when using requestSubmit

https://github.com/WICG/webcomponents/issues/814

packages/components/src/components/button-pure/button-pure.tsx
kaiszybiak: updated

packages/components/src/components/button-pure/button-pure.tsx
kaiszybiak: updated as discussed

kaiszybiak: Actually the optional chaining was due to:

 ❯ tests/unit/specs/element-internals.spec.ts (1 test | 1 failed) 14ms
   × Element Internals > should ensure both attachInternals and this.internals use optional chaining (?.) only

packages/components/CHANGELOG.md
```diff
@@ -30,6 +30,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0),
 - Styles: support for `vanilla-extract` available under
   `import { … } from '@porsche-design-system/components-{js|angular|react|vue}/styles/vanilla-extract';`
   ([#3666](https://github.com/porsche-design-system/porsche-design-system/pull/3666))
+- `Button`, `Button-Pure`: Add `form` prop to explicitly associate the component with a form, even when it's not
```
henrijoss, line 33: Move to unreleased

henrijoss (approved):

PR #3651, opened 2024-11-28
Show "with image" option in select input field (Issue/3549)

### LINKED ISSUE #3549: Show "with image" option in select input field
Scope
---
Display the selected option WITH the image in the select input field not only in the options list.

Current implementation:
https://github.com/user-attachments/assets/f49c5f77-001d-4c78-a716-d75690d67686

Expected:
Bildschirmfoto 2024-10-22 um 10 40 15

Acceptance criteria
---
- [x] slotted image is visible in the select inputfield after selection is made
- [ ] Component is aligned in Figma

Subtasks
---
- [ ] Task

o-hook:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3549/components/select/examples

#### Scope

#### Resolved Issue

Resolves #3549


packages/components/src/components/select/select/select.tsx
```diff
@@ -151,6 +151,7 @@ export class Select {
   private isNativePopoverCase: boolean = false;
   private parentTableElement: HTMLElement;
   private popoverElement: HTMLElement;
+  private hasSlottedImage: boolean = false;
```
henrijoss, line 154: Why not directly save the current img src here and reuse in the render instead of calling `this.getSelectedOptionImagePath` again?

packages/components/src/components/select/select/select.tsx
```diff
@@ -344,6 +354,7 @@ export class Select {
     }
     this.updateMenuState(false);
     this.combobox.focus();
+    this.hasSlottedImage = this.getSelectedOptionImagePath(this.selectOptions) !== '';
```
henrijoss, line 357: Maybe it would be safer to update this in the value watcher? Test in E2E if dynamic value change would update the image

packages/components-js/tests/e2e/specs/select.e2e.ts
henrijoss, line 1: We could also add a test when setting the value programmatically e.g. setProperty

henrijoss (approved):

PR #3763, opened 2025-02-24
[FEATURE]: Create `p-input-password` (issue/3389)

### LINKED ISSUE #3389: Create `p-input-password`
Scope
---
Currently, we're providing `p-textfield-wrapper` which requires an `input type="password"` as child element. To provide a more solid/modern variant we want to create a new component called p-input-password with an `input type="password"` within the Shadow DOM.

Notes  
---
- See `p-textarea` / `p-checkbox` for reference
- Use `formAssociated / attachInternals` API https://stenciljs.com/docs/form-associated 
- Can be implemented as either a controlled or an uncontrolled component. Challenge how `onChange` / `onInput` is working in React vs Vanilla JS
- Ensure native attributes like e.g. `required` don't interfere with custom prop `required`
- Besides interface proposal, check if other native `input` attributes should be exposed as `p-input-password` component prop, https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attributes
- Interface proposal:
```
# Name: 
p-input-password

# Props:
name: string;
value?: string = '';
auto-complete?: 'off' | 'on' | '' = '';
disabled?: boolean = false;
form?: string;
max-length?: number;
min-length?: number; 
pattern? string;
placeholder?: string = '';
read-only?: boolean = false;
required?: boolean = false;

loading?: boolean = false; // experimental prop, check if real use cases exist before implementing

label?: string = '';
hideLabel?: boolean | BreakpointCustomizable<boolean> = false;
description?: string = '';
message?: string = '';
state?: 'none' | 'error' | 'success' = 'none';
theme?: Theme = 'light';

# Events:
blur
change
input

# Slots:
- slot="label"
- slot="description"
- slot="message"
```

Acceptance criteria
---
- [ ] `p-input-password` is provided by npm package `@porsche-design-system/components-{js|angular|react|vue}` with the same feature scope like `p-textfield-wrapper > input type="password"` and supports the most important native `input` attributes
- [ ] `p-input-password` supports toggle password visibility button
- [ ] `p-input-password` supports compact mode
- [ ] `p-input-password` is documented at https://designsystem.porsche.com
- [ ] ~~`p-textfield-wrapper` is deprecated~~

Subtasks
---
- [ ] Task

kaiszybiak:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3389/

#### Scope

- Introduce new input password component
- Split `label` into `label` and `legacy-label`
- Added `docs/coding-standards-and-guidelines.md`

#### Resolved Issue

Resolves #3389


packages/components/src/components/input-password/input-password-styles.ts
```diff
@@ -0,0 +1,183 @@
...
+      ':host': {
+        display: 'block',
+        ...addImportantToEachRule({
+          ...colorSchemeStyles,
+          ...hostHiddenStyles,
+        }),
+      },
+      ...preventFoucOfNestedElementsStyles,
+
+      input: {
```
henrijoss, line 83: Maybe `unset: all`

packages/components/src/components/input-password/input-password-styles.ts
```diff
@@ -0,0 +1,183 @@
...
+        boxSizing: 'content-box',
+        paddingInlineStart: `var(${cssVariableInputPaddingStart})`, // iOS Safari 14.5 can't handle padding-inline shorthand with css variables
+        paddingInlineEnd: `var(${cssVariableInputPaddingEnd})`, // iOS Safari 14.5 can't handle padding-inline shorthand with css variables
+      },
+      ':not(input[type="password"])': {
+        textOverflow: 'ellipsis',
+      },
+      ...(hoverMediaQuery({
+        // with the media query the selector has higher priority and overrides disabled styles
+        'input:not(:disabled):not(:focus):not([readonly]):hover,label:hover~.wrapper input:not(:disabled):not(:focus):not([readonly])':
```
henrijoss, line 114: Maybe this can be improved now

packages/components/src/components/input-password/input-password-utils.ts
```diff
@@ -0,0 +1,9 @@
+import type { FormState } from '../../utils/form/form-state';
+export type InputPasswordState = FormState;
+
+export const INPUT_PASSWORD_AUTO_COMPLETE = ['off', 'on', 'current-password', 'new-password', ''] as const;
```
henrijoss, line 4: Check switch between type text and type password

packages/components/src/components/input-password/input-password.tsx
```diff
@@ -0,0 +1,276 @@
...
+            aria-describedby={`${descriptionId} ${messageId}`}
+            aria-invalid={this.state === 'error' ? 'true' : null}
+            id={id}
+            ref={(el: HTMLInputElement) => (this.inputElement = el)}
+            onInput={this.onInput}
+            onChange={this.onChange}
+            onBlur={this.onBlur}
+            name={this.name}
+            form={this.form}
+            type={this.showPassword ? 'text' : 'password'}
```
henrijoss, line 224: Check if these types differ in any way other than visually

packages/components/src/components/input-password/input-password.examples.md
```diff
@@ -0,0 +1,159 @@
...
+
+  get labelMarkup() {
+    return `<p-input-password name="some-name" label="Some label"></p-input-password>`;
+  }
+
+  
+  typePasswordWithoutPasswordToggle =
+`<p-text-field-wrapper label="Some label" show-password-toggle="false">
+  <input type="password" name="some-name" value="some password" />
+</p-text-field-wrapper>`;
```
o-hook, line 125: Shouldn't this be the new `p-input-password` component?

packages/components/src/components/input-password/input-password.examples.md
```diff
@@ -0,0 +1,159 @@
...
+</Notification>
+
+<TableOfContents></TableOfContents>
+
+## Basic example
+
+Instead of relying on slotted content, the `p-input-password` component offers a `value` attribute and property that
+remain synchronized with user input.
+
+<Playground :markup="labelMarkup" :config="config"></Playground>
```
o-hook, line 30: In my Chrome Browser (v133) the input element doesn't get focused (blinking cursor) if i click on the label. All other Browsers work fine. Can you reproduce that?

packages/components/src/components/input-password/input-password.tsx
kaiszybiak: I think it is the way to go when implementing password visibility. It is also provided as an example here -->  https://www.w3schools.com/howto/howto_js_toggle_password.asp.

Besides that will likely have the following differences:

**Autocomplete Behavior**

- `password`: The browser may offer to save or autofill passwords.
- `text`: Usually does not trigger password managers.

**Clipboard & Accessibility**

- Some browsers may prevent copying text from a password field.
- Screen readers may treat password fields differently for security.

packages/components/src/components/input-password/input-password-utils.ts
kaiszybiak: Depending on the Browser and password manager 'current-password' and 'new-password' might be ignored if `type="text"`. However in Chrome and with 1Password as password manager it autocomplete feature worked also with password visibility on.

packages/components/src/components/input-password/input-password-styles.ts
```diff
@@ -0,0 +1,183 @@
...
+        paddingInline,
+        width: 'auto',
+        height,
+        boxSizing: 'content-box',
+        paddingInlineStart: `var(${cssVariableInputPaddingStart})`, // iOS Safari 14.5 can't handle padding-inline shorthand with css variables
+        paddingInlineEnd: `var(${cssVariableInputPaddingEnd})`, // iOS Safari 14.5 can't handle padding-inline shorthand with css variables
+      },
+      ':not(input[type="password"])': {
+        textOverflow: 'ellipsis',
+      },
```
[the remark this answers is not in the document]
kaiszybiak: Bildschirmfoto 2025-03-04 um 10 33 31

Bildschirmfoto 2025-03-04 um 10 33 42

I think this is fine

packages/components/src/components/input-password/input-password.tsx
```diff
@@ -0,0 +1,276 @@
...
+  @Prop() public hideLabel?: BreakpointCustomizable<boolean> = false;
+
+  /** @experimental Show or hide password toggle for `input type="password"`. */
+  @Prop() public showPasswordToggle?: boolean = true;
+
+  /** Adapts the color depending on the theme. */
+  @Prop() public theme?: Theme = 'light';
+
+  /** Emitted when the password input loses focus after its value was changed. */
+  @Event({ bubbles: true }) public change: EventEmitter<InputPasswordChangeEventDetail>;
```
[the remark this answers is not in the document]
kaiszybiak: I think we are aligned with native event bubbling: https://developer.mozilla.org/en-US/docs/Web/API/Element/blur_event 

> The blur event fires when an element has lost focus. The event does not bubble

packages/components/src/components/input-password/input-password-styles.ts
kaiszybiak: updated

packages/components/src/components/input-password/input-password-styles.ts
kaiszybiak: `all: reset`  might reset more than intended. (https://developer.mozilla.org/en-US/docs/Web/CSS/unset)

packages/components/src/components/input-password/input-password.examples.md
kaiszybiak: Good catch! I think the issue was `shadow: { delegatesFocus: true }`. We have similar issues for `p-textarea` and `p-pin-code`.

packages/components/src/components/input-password/input-password.examples.md
kaiszybiak: updated

packages/components/src/components/input-password/input-password.tsx
```diff
@@ -0,0 +1,278 @@
...
+  /** The validation state. */
+  @Prop() public state?: InputPasswordState = 'none';
+
+  /** The message styled depending on validation state. */
+  @Prop() public message?: string = '';
+
+  /** Show or hide label and description text. For better accessibility it is recommended to show the label. */
+  @Prop() public hideLabel?: BreakpointCustomizable<boolean> = false;
+
+  /** @experimental Show or hide password toggle for `input type="password"`. */
```
henrijoss, line 127: Why is this experimental?

packages/components/src/components/input-password/input-password.tsx
```diff
@@ -0,0 +1,278 @@
...
+  // In the React wrapper, all props are synced as properties on the element ref, so reflecting "name" as an attribute ensures it is properly handled in the form submission process.
+
+  /** The password input value. */
+  @Prop({ mutable: true }) public value?: string = '';
+
+  /** Specifies whether the input can be autofilled by the browser */
+  @Prop() public autoComplete?: InputPasswordAutoComplete = '';
+
+  /** Specifies whether the input should have its spelling and grammar checked */
+  @Prop() public spellCheck?: boolean;
```
henrijoss, line 95: Does this make sense for a password?

packages/components-js/tests/e2e/specs/input-password.e2e.ts
```diff
@@ -0,0 +1,463 @@
...
+      props: { name: 'Some name', state: 'error' },
+      useSlottedLabel: true,
+      useSlottedMessage: true,
+      useSlottedDescription: true,
+    });
+    const status = await getLifecycleStatus(page);
+
+    expect(status.componentDidLoad['p-input-password'], 'componentDidLoad: p-input-password').toBe(1);
+    expect(status.componentDidLoad['p-icon'], 'componentDidLoad: p-icon').toBe(2);
+
```
henrijoss, line 425: Check for `p-button-pure` is missing

packages/components-js/tests/e2e/specs/input-password.e2e.ts
henrijoss, line 1: Consider adding a test for the password toggle functionality. Also thing about everything which could change dynamically and check that it is working dynamically like e.g. `disabled`

packages/components/src/components/input-password/input-password.examples.md
```diff
@@ -0,0 +1,157 @@
...
+<a href="https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals">ElementInternals</a> API, which has limited
+browser support. For broader browser compatibility, consider using a
+<a href="components/input-password/examples#controlled">controlled</a> approach instead.
+</Notification>
+
+<TableOfContents></TableOfContents>
+
+## Basic example
+
+Instead of relying on slotted content, the `p-input-password` component offers a `value` attribute and property that
```
henrijoss, line 27: Slotted content would not make sense here since its similar to the native `input`

packages/components/src/components/input-password/input-password.tsx
```diff
@@ -0,0 +1,278 @@
...
+            </PrefixedTagNames.pButtonPure>
+          )}
+        </div>
+        <StateMessage state={this.state} message={this.message} theme={this.theme} host={this.host} />
+      </div>
+    );
+  }
+
+  private onChange = (e: Event): void => {
+    this.change.emit(e);
```
henrijoss, line 258: Check if we need to prevent native event bubbling

packages/components/src/components/input-password/input-password-styles.ts
```diff
@@ -0,0 +1,172 @@
...
+  formElementPaddingVertical,
+} from '../../styles/form-styles';
+import type { BreakpointCustomizable, Theme } from '../../types';
+import { getCss } from '../../utils';
+import type { FormState } from '../../utils/form/form-state';
+import { getFunctionalComponentLabelStyles } from '../common/label/label-styles';
+import { getFunctionalComponentStateMessageStyles } from '../common/state-message/state-message-styles';
+
+export const cssVariableInputPaddingStart = '--p-internal-input-password-input-padding-start';
+export const cssVariableInputPaddingEnd = '--p-internal-input-password-input-padding-end';
```
henrijoss, line 32: Why are these css variables?

kaiszybiak: removed

packages/components/src/components/input-password/input-password.tsx
kaiszybiak: removed

packages/components/src/components/input-password/input-password.tsx
kaiszybiak: removed

packages/components-js/tests/e2e/specs/input-password.e2e.ts
kaiszybiak: updated tests

packages/components/src/components/input-password/input-password-styles.ts
```diff
@@ -0,0 +1,167 @@
...
+      '&:has(input:focus:not([readonly]))': {
+        borderColor: primaryColor,
+        ...prefersColorSchemeDarkMediaQuery(theme, {
+          borderColor: primaryColorDark,
+        }),
+      },
+      ...hoverMediaQuery({
+        ...(!disabled &&
+          !readOnly && {
+            '&:hover:not(:has(input:focus)):not(:has(p-button-pure:hover))': hoverStyles,
```
stephanschroeter, line 125: You shouldn't use `p-button-pure` as selector since it won't work with prefixed components, use css class instead.

stephanschroeter, line 125: Instead of repeating `:not()` you can write `:not(… , …)`

stephanschroeter (dismissed):

henrijoss (approved):

PR #3796, opened 2025-03-19
Improve aria-live announcements of state messages of form components (issue/3769)

### LINKED ISSUE #3769: Improve `aria-live` announcements of state messages of form components
Scope
---
The `error` and `success` states of form components are getting rendered visually if the related props are set. To provide this visual change also for screen reader users, those messages have the corresponding ARIA roles of `alert` and `status`. The problem with such roles is, that those elements must be **initially** rendered, so that they can expose their content changes to assistive technologies. Currently state messages are **conditionally** rendered which means that not all screen readers announce those states automatically. 

Notes  
---
- The states are provided by the `StateMessage`functional component
- To provide the broadest screen reader support, the ARIA live-region container must be rendered initially (visually hidden) and only the contents and visibility should change.

Acceptance criteria
---
- [ ] All contents and content changes of state messages must be exposed by screen readers

Subtasks
---
- [ ] Task

o-hook:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3769/

#### Scope

#### Resolved Issue

Resolves #3769


packages/components/src/components/common/state-message/state-message.tsx
```diff
@@ -14,18 +14,19 @@ export const messageId = 'message';
 export const StateMessage: FunctionalComponent<StateMessageProps> = ({ state, message, theme, host }) => {
   const PrefixedTagNames = getPrefixedTagNames(host);
   const isErrorState = state === 'error';
+  const isSuccessState = state === 'success';
```
stephanschroeter, line 17: to be in sync with other conditions in state-message, we could simply use `isErrorState` instead of `isSuccessState`

packages/components/src/components/common/state-message/state-message.tsx
```diff
@@ -16,16 +16,16 @@ export const StateMessage: FunctionalComponent<StateMessageProps> = ({ state, me
   const isErrorState = state === 'error';
 
   return (
-    hasMessage(host, message, state) && (
-      <span id={messageId} class="message" role={getRole(state)}>
+    <span id={messageId} class="message" role={state === 'success' ? 'status' : 'alert'}>
```
stephanschroeter, line 19:
```suggestion
    {/* needs to be rendered always to have correct behaviour for screen readers */}
    <span id={messageId} class="message" role={state === 'success' ? 'status' : 'alert'}>
```

stephanschroeter (approved):

PR #3802, opened 2025-03-21
chore(docker): Check semantic version before release

kaiszybiak:
### Pull Request Checklist

#### References

#### Scope

Verify version bump is valid before release

#### Resolved Issue

Resolves [IssueNumber]

### DIFF
--- a/docker/bin/run-check-version
+++ b/docker/bin/run-check-version
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[0;33m'
+NC='\033[0m'
+
+PACKAGE_LOCATION="packages/components"
+PACKAGE_JSON="${PACKAGE_LOCATION}/package.json"
+PACKAGE_VERSION=$(grep version "${PACKAGE_JSON}" | head -n 1 | sed -E 's/.*"version": "(.*)",/\1/')
+
+# Check if the new version is provided
+if [[ -z "${1}" ]]; then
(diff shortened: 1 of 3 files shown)


docker/bin/run-prepare-release-assets
```diff
@@ -3,10 +3,9 @@
 set -o errexit
 set -o pipefail
 
-if [[ -z "${1}" ]]; then
-  echo "Please provide a target version, e.g. './docker.sh run-prepare-release-assets 1.2.3'"
-  exit 1
-fi
+SCRIPT_DIR="$(cd `dirname ${0}` && pwd)"
+
+source "${SCRIPT_DIR}/run-check-version"
```
stephanschroeter, line 8: I think run-check-version will only check for component's package.json?

docker/bin/run-check-version
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
kaiszybiak: Yes, I will do a docker image release after everything else is clarified.

docker/bin/run-prepare-release-assets
kaiszybiak: Yes, when exec `./docker.sh run-prepare-release-components ${TARGET_VERSION}` it will compare the provided `TARGET_VERSION` with the version of `components`, which should be enough as the versions of:

- `components`
- `components-js/projects/components-wrapper`
- `components-js`
- `components-angular`
- `components-react`
- `components-vue`
etc.

are all in sync and will all be set to `TARGET_VERSION`.

henrijoss (approved):

PR #3819, opened 2025-04-01
Refactor Checkbox of Multi Select (Issue/3193)

### LINKED ISSUE #3193: [BUG] - AXE-Core test fails if tested with open dropdown of `Multi Select`
### JS Framework

all

3.15.0-rc.1

all

all

### Browser version

all

1. Go to https://designsystem.porsche.com/v3/components/multi-select/examples
2. Open a MultiSelect
3. Do partial test with AXE DevTools
4. See error: Ensures elements with an ARIA role that require child roles contain them

### Expected behavior

AXE-Core test should be also valid for dropdown of `Multi Select`
Re-enable Axe Core Tests for multi-select

### Additional information

The AXE-Core test fails due to the `role="status"` of the loading message inside of the checkbox component which is a child element of the `role="listbox"` element which is not allowed in this context.

Bildschirmfoto 2024-04-22 um 11 17 11

**A11y tree:**
Bildschirmfoto 2024-04-22 um 11 20 31

o-hook:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3193/

#### Scope

#### Resolved Issue

Resolves #3193


packages/components/CHANGELOG.md
```diff
@@ -14,6 +14,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0),
 
 ### [Unreleased]
 
+#### Fixed
+
+- `Multi Select`: refactor checkbox and fix A11Y test
```
stephanschroeter, line 19: I don't think it's relevant for the consumer that we've fixed an a11y test.

stephanschroeter, line 19:
```suggestion
- `Multi Select`: enable full a11y compliance
```

packages/components/src/components/multi-select/multi-select-option/multi-select-option.tsx
```diff
@@ -39,16 +38,14 @@ export class MultiSelectOption {
   public render(): JSX.Element {
     validateProps(this, propTypes);
     const { theme = 'light', selected, highlighted } = this.host;
-    attachComponentCss(this.host, getComponentCss, theme);
-
-    const PrefixedTagNames = getPrefixedTagNames(this.host);
     const isDisabled = this.disabled || this.host.disabledParent;
 
+    attachComponentCss(this.host, getComponentCss, theme, isDisabled, selected);
```
stephanschroeter, line 43: To be in sync with param naming, we should use `isSelected`.

o-hook: But in this case, the `selected` refers to the native `selected`, so I can't change the destructured const.

packages/components/src/components/multi-select/multi-select-option/multi-select-option-styles.ts
```diff
@@ -13,11 +60,53 @@ export const getComponentCss = (theme: Theme): string => {
...
+        display: 'block',
+        paddingTop,
+      },
+    },
+    option: {
+      ...getOptionJssStyle('multi-select-option', 1, theme),
+      columnGap: '8px',
     },
-    option: getOptionJssStyle('multi-select-option', 1, theme),
     checkbox: {
```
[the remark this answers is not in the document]
o-hook: Maybe parts of the styles, because the fake Checkbox doesn't need all of the styles, e.g. `compact` styles or `indeterminate`, `loading`, etc. 
Or we make a configurable styles function...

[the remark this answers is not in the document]
o-hook: Some styles are shared now.

packages/components/src/styles/checkbox/checkbox-styles.ts
```diff
@@ -0,0 +1,154 @@
...
+import { getCheckboxCheckedBaseStyles } from './checkbox-checked-base-styles';
+import { getHighContrastColors, getThemedColors } from '../colors';
+
+export const getIndeterminateSVGBackgroundImage = (fill: string): string => {
+  return getInlineSVGBackgroundImage(`<path fill="${fill}" d="m20,11v2H4v-2h16Z"/>`);
+};
+
+export const cssVarInternalCheckboxScaling = '--p-internal-checkbox-scaling';
+
+export const getCheckboxStyles = (
```
stephanschroeter, line 20: I would move it back to local checkbox styles.

o-hook: Done

Issue #3864, opened 2025-05-21
Vue Provider does not work when initially set to `dark` or `auto`


henrijoss:
The `PorscheDesignSystemProvider` does not work in Vue when the `theme` is set to `dark` or `auto` initially. Dynamically switching from `light` to `dark` or `auto` does works.

### JS Framework

Vue

### Porsche Design System version

3.28.0-rc.2

### Reproduction Example

https://stackblitz.com/edit/jnwebgem?file=package.json

### Steps to reproduce the behavior

1. Go to https://stackblitz.com/edit/jnwebgem?file=package.json
2. See `PorscheDesignSystemProvider` `theme` is set to `dark` initially
3. Button is still in light mode

### Expected behavior

The theme is provided to all components correctly when setting the initial theme to `dark` or `auto`

### Notes

- To avoid confusion: Stackblitz has a Bug where `@media (prefers-color-scheme: dark)` is not working

stephanschroeter:
- Fixed in v3
- Not needed anymore in v4 since native `color-scheme` is used instead

PR #3898, opened 2025-06-24
[Fix]: Vue Provider does not work when initially set to dark or auto (#3864)

kaiszybiak:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3864/

#### Scope

This change ensures the provider applies the initial `theme` correctly and keeps downstream components in sync when `theme` changes, without re-initializing the design system at runtime.

##### Key changes:

- Moved `load(props)` into `onMounted` to guarantee one-time initialization with the correct initial props.
- Initialized `themeRef` from `props.theme` immediately in setup.
- Added a `watch` on `props.theme` to update `themeRef` reactively, so injected consumers respond to theme changes.
- Removed any use of `onUpdated` for theme handling to avoid unnecessary or redundant updates.

#### Resolved Issue

Resolves [#3864]


packages/components-vue/projects/vue-wrapper/src/PorscheDesignSystemProvider.vue
```diff
@@ -12,19 +12,21 @@ type Props = {
...
   theme: 'light',
 });
 
-onUpdated(() => {
-  themeRef.value = props.theme;
+const themeRef = ref<Theme>(props.theme);
+
+onMounted(() => {
+  load(props); // runtime prefix or cdn change is not supported
+});
```
henrijoss: `load()` should be called asap since it is loading the core chunk and should not wait until `onMounted`

packages/components-vue/projects/vue-wrapper/tests/unit/specs/porsche-design-system-provider.spec.ts
henrijoss, line 1: Add a test to `packages/components-vue/tests/e2e/specs/provider.e2e.ts` checking that the components are initialised with the correct theme

packages/components-vue/tests/e2e/specs/provider.e2e.ts
henrijoss, line 1: Won't the theme be initialised with `light` and then the query param watcher changes it to `dark`? Can we create an isolated test where we don't have to rely on our implementation of the app?

kaiszybiak: The `themeRef` is initialized from `props.theme`, which is already resolved when `<script setup>` executes. That means when `PorscheDesignSystemProvider` is mounted with `:theme="'dark'"`, ` themeRef.value` is immediately 
`'dark'`.
The `watch()` is only there to support reactive updates later. No transient `'light'` value is ever provided to child components.

henrijoss: Maybe we can add three Providers with hardcoded values for the themes and check them directly.

```
    <PorscheDesignSystemProvider cdn="auto" :theme="'dark'">
      <ComponentToTest />
    </PorscheDesignSystemProvider>
```

Nested Providers should also work.

kaiszybiak: updated

packages/components-vue/projects/vue-wrapper/tests/unit/specs/porsche-design-system-provider.spec.ts
```diff
@@ -127,3 +127,45 @@ describe('PorscheDesignSystemProvider', () => {
...
+            </PorscheDesignSystemProvider>
+          `,
+      })
+    );
+
+    const text = wrapper.find('[data-test="theme"]').text();
+    expect(text).toBe(theme);
+  });
+
+  test('nested providers override parent theme', () => {
```
henrijoss, line 158:
```
<PorscheDesignSystemProvider cdn="auto" theme="dark">
  <ThemeConsumer />
</PorscheDesignSystemProvider>
```

One Provider should be enough. Maybe add the test three times for `light`, `dark` & `auto`. Also render some components e.g. button and check correct theme.

packages/components-vue/src/App.vue
henrijoss, line 1: Instead of all of this logic we should keep it simple and just have a test with the Provider already included and hardcoded theme value.

packages/components-vue/projects/vue-wrapper/tests/unit/specs/porsche-design-system-provider.spec.ts
kaiszybiak: updated

henrijoss (dismissed):

stephanschroeter (approved):

Issue #3537, opened 2024-10-14
Add filter mode to p-multi-select


marcelbertram:
Scope
---
Filter mode is added to p-multi-select

Notes  
---
- Figma: tbd

Acceptance criteria
---
- [ ] Filter mode is added to p-multi-select

Subtasks
---
- [ ] Task

henrijoss:
Done in https://github.com/porsche-design-system/porsche-design-system/pull/3953

PR #3943, opened 2025-07-22
feat(components): align multi-select with select | hj | #3537

henrijoss:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3537

#### Scope

#### Resolved Issue

Resolves [IssueNumber]


packages/components/src/components/multi-select/multi-select/multi-select.tsx
```diff
@@ -311,16 +297,43 @@ export class MultiSelect {
...
+            class="filter"
+            name="filter"
+            label="Filter options"
+            hideLabel={true}
+            autoComplete="off"
+            clear={true}
+            indicator={true}
+            compact={true}
+            theme={this.theme}
+            {...getComboboxFilterAriaAttributes()}
```
o-hook: Sprouting `aria` attributes directly on host element has no effect. We have to discuss how we can manage this, e.g. by not having shadowDom in this scenario.

We have to place the needed `aria` attributes directly on the `<input/>` element and also manage `aria-activedescendant` with elementInternals.

o-hook (changes requested):

Issue #3486, opened 2024-08-30
[FEATURE] Add slot="button" to p-popover to make usage more generic


stephanschroeter:
Scope
---
We need to enable the popover component for contexts without "i" to apply for other buttons and slots. To better support unified applications of different patterns we should think of introducing a header and footer slot 

Out of Scope
---
List topics that are not part of this task.

Notes  
---
- Needed for AI Annotation: https://github.com/porsche-design-system/digital-design/issues/184

Acceptance criteria
---
- [ ] Criteria

Subtasks
---
- [ ] Task

PR #3976, opened 2025-08-15
fix(components): popover is handled correctly for controlled/uncontro…

henrijoss:
…lled, with native support/without | hj | #3486-fix

### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3486-fix

#### Scope

#### Resolved Issue

https://designsystem.porsche.com/issue/3486
- Multiple clicks on popover button in controlled case caused state to be falsly synced
- Outside click closed popover in controlled state without dismiss event being necessary

#### Improvements

- Opening/closing is handled automatically regardless of custom button
- Document click listener is only used for browsers not supporting native popover
- Uses native popover in "auto" mode for outside click

### DIFF
--- a/packages/components-react/projects/react-ssr-wrapper/tests/unit/specs/__snapshots__/react-ssr-wrapper.spec.tsx.snap
+++ b/packages/components-react/projects/react-ssr-wrapper/tests/unit/specs/__snapshots__/react-ssr-wrapper.spec.tsx.snap
@@ -18120,6 +18120,9 @@ exports[`should render dsr component for PPopover 1`] = `
   :host([hidden]) {
     display: none !important;
   }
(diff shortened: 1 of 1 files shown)


packages/components-react/projects/react-ssr-wrapper/tests/unit/specs/__snapshots__/react-ssr-wrapper.spec.tsx.snap
```diff
@@ -18205,7 +18208,6 @@ exports[`should render dsr component for PPopover 1`] = `
   }
     </style>
     <button
-      aria-expanded="false"
```
stephanschroeter, line 18208: intended?

Issue #4000, opened 2025-08-25
Expose native `onChange` event in `p-checkbox`


henrijoss:
Scope
---
The `p-checkbox` currently exposes an `onUpdate` event, even though it internally uses a native `input` of type checkbox. To better align with its native counterpart and simplify form integration, it should instead expose the native `onChange` event.

Out of Scope
---
-

Notes  
---
-

Acceptance criteria
---
- [ ] native `onChange` event is exposed
- [ ] `onUpdate` event is marked as depracated

Subtasks
---
- [ ] Task

PR #4010, opened 2025-08-27
Expose native onChange event in p-checkbox (#4000)

kaiszybiak:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/4000/

#### Scope

#### Resolved Issue

Resolves [#4000]


packages/storefront/src/app/news/migration-guide/porsche-design-system/page.mdx
henrijoss, line 1: This is the migration guide from v2 to v3

kaiszybiak: reverted

packages/components/src/components/checkbox/checkbox.tsx
```diff
@@ -92,21 +97,27 @@ export class Checkbox {
...
   @Prop() public theme?: Theme = 'light';
 
-  /** Emitted when checkbox checked property is changed. */
+  /**
+   * Emitted when checkbox checked property is changed.
+   * @deprecated since v3.30.0, will be removed with next major release, use `change` event instead.
+   */
   @Event({ bubbles: false }) public update: EventEmitter<CheckboxUpdateEventDetail>;
 
+  /** Emitted when the checkbox loses focus after its value was changed. */
```
henrijoss: It doesn't need to lose focus to emit the change event.

From the MDN docs:
When a [<input type="checkbox">](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/checkbox) element is checked or unchecked (by clicking or using the keyboard);

packages/components/CHANGELOG.md
```diff
@@ -20,6 +20,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0),
   `aria` prop to provide additional context for screen readers.
   ([#3969](https://github.com/porsche-design-system/porsche-design-system/pull/3969))
 
+### Changed
```
henrijoss: Shouln't this be under the Added section and mention that the `change` event was added?

packages/components-js/tests/e2e/specs/checkbox.e2e.ts
```diff
@@ -779,13 +785,18 @@ test.describe('Event', () => {
...
+      expect((await getEventSummary(host, 'change')).counter).toBe(0);
 
       await input.click();
       await waitForStencilLifecycle(page);
 
       expect((await getEventSummary(host, 'update')).counter).toBe(1);
       expect((await getEventSummary(host, 'update')).details).toEqual([{ checked: true, name: 'some-name', value }]);
+
+      expect((await getEventSummary(host, 'change')).counter).toBe(1);
+      expect((await getEventSummary(host, 'change')).details).toEqual([{ isTrusted: true }]);
```
henrijoss, line 799: Shouldn't this check if the target or checked prop is there?

kaiszybiak: If I do `expect((await getEventSummary(host, 'change')).targets).toEqual([...]);` I most likely have to extent our `addEventListener` function because otherwise I get the error `TypeError: e.cloneNode is not a function`. As an alternative instead of forwarding the native event (`this.change.emit(e);`) I could do ` this.change.emit({ value: this.value, name: this.name, checked });` but for the other input components we also forward the native event, so I would suggest leave it as it is or even remove this line.

packages/components/CHANGELOG.md
kaiszybiak: updated

packages/components/src/components/checkbox/checkbox.tsx
kaiszybiak: updated

packages/components-js/tests/e2e/specs/checkbox.e2e.ts
henrijoss: The details should also contain `target.checked`?

kaiszybiak: To stay consistent with other input components, we’re forwarding the native event, so details only has native properties like `isTrusted`. Including `target.checked` would require changing all inputs to emit a normalized object instead of the native event.

stephanschroeter (approved):

PR #4046, opened 2025-10-06
Fix Select-Wrapper Styles & Add Change Event Select/Multi-Select

henrijoss:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3999

#### Scope

#### Resolved Issue

Resolves [IssueNumber]


packages/components/src/components/multi-select/multi-select/multi-select.tsx
```diff
@@ -529,13 +543,18 @@ export class MultiSelect {
   };
 
   private emitUpdateEvent = (): void => {
+    this.change.emit({
```
kaiszybiak, line 546: Don't we want to emit the native change `Event` like we did for the Checkbox ? (https://github.com/porsche-design-system/porsche-design-system/pull/4010)

henrijoss: There is no native change event available without added a hidden synchronized `select`. Since we already have the `value` and `name` emitted, it't not really necessary

Issue #3972, opened 2025-08-14
`<p-flag />` follow up


stephanschroeter:
Scope
---
Thread: https://porschedev.slack.com/archives/CT7AVHTTQ/p1761041555226159

Notes  
---
- International Flag: https://github.com/porsche-home-nav/country-selector/blob/e7100d93fcb6239dec0fa93e984d08f6eca749d6/src/assets/img/flags/International.svg
- Flags for `AD, GI, LI, MC, ME, TW`: https://github.com/porsche-home-nav/country-selector/tree/e7100d93fcb6239dec0fa93e984d08f6eca749d6/src/assets/img/flags/other
- Ensure all SVG flag's have the same viewbox
- Check for what `xx` stands for in ISO 3166-1 alpha-2 country code
- Don't change the default flag, it's still "de"

Acceptance criteria
---
- [ ] Component is available in Figma
- [ ] Generic flag (`name="xx"`) is implemented 
- [ ] Flags for `AD, GI, LI, MC, ME, TW` are implemented
- [ ] When an invalid country name is used then the fallback flag `xx` is used

Subtasks
---
- [ ] Task

kaiszybiak:
fyi: using `xx` is a widely accepted convention. However, it's important to note that this usage is not officially recognized by `ISO 3166-1.`

PR #4074, opened 2025-10-22
[Feat]: `<p-flag />` follow up

kaiszybiak:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3972/

#### Scope

#### Resolved Issue

Resolves [#3972]


packages/components/src/components/flag/flag-utils.spec.ts
```diff
@@ -17,6 +18,6 @@ describe('buildFlagUrl()', () => {
 
   it('should return default flag url if flag name is not in manifest', () => {
```
stephanschroeter, line 19:
```suggestion
  it('should return unknown flag url if flag name is not in manifest', () => {
```

stephanschroeter (approved):

packages/components/src/components/flag/flag-utils.spec.ts
kaiszybiak: will do this change in https://github.com/porsche-design-system/porsche-design-system/pull/4073 due to flaky pipeline.

PR #4078, opened 2025-10-23
[Bug] - Radio Group not focusable in initial (deselected) state (Issue/4062)

### LINKED ISSUE #4062: [Bug] - `Radio Group` not focusable in initial (deselected) state, some other ARIA issues
Description
---

**Keyboard issue:**
The `Radio Group` component is not keyboard accessible (focusable) if none of the items has a `checked` state. If no item is selected, then the first one needs to be focusable.

**SEMANTICS/ARIA issues:**
- A radio group must be wrapped inside a `fieldset` element with a `legend` as the overarching label (and not with a `label` element). The `aria-labelledby` attribute has a wrong value (and can probably removed if a `fieldset` and `legend` is used.
- If the `Radio Group` is in `error` state, the error message is not announced if a radio button has focus. This is because the `aria-labelledby` attribute has no reference to the ID of the error message (cross shadowDOM issue).
- `aria-describedby` attribute of the radio-group is missing. This should point to the ID of the overarching description.
- Screenreader announcement of number of options not correct: "Option one of one" (instead of sum of all options)
Image

Notes  
---
See also official documentation: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/radio

Acceptance criteria
---
- [ ] `Radio Group` is fully keyboard accessible
- [ ] Overarching label of `Radio Group` is semantically correct

Subtasks
---
- [ ] Task

o-hook:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/4062/

#### Scope

#### Resolved Issue

Resolves #4062


packages/components/src/components/radio-group/radio-group/radio-group-styles.ts
```diff
@@ -83,6 +83,9 @@ export const getComponentCss = (
       display: 'grid',
       justifySelf: 'flex-start',
       rowGap: spacingStaticXSmall,
+      margin: 0,
+      padding: 0,
+      border: 'none',
```
henrijoss: Could also use `all: unset`

henrijoss (dismissed):

henrijoss (dismissed):

henrijoss (approved):

Issue #3988, opened 2025-08-20
[FEATURE] - onInputChange onInputEnter event in p-select component


JosseArturo:
### Portfolio

No Portfolio

### ART and Team

Brake development

### Contact Details

igor.iroz.asin@porsche.de

### Problem

Due to many possible options inside the select dropdown menu it is required to have a backend call, that brings just some options to show in the select menu, once the user type on the filter (input text) inside the  p-select component

### Solution ideas and examples

Add onInputChange and onInputEnter event triggered by the interaction with the search (input filter) embedded in the select component (once the filter Property is activated)

### Business priority

None

henrijoss:
Implemented with: https://github.com/porsche-design-system/porsche-design-system/pull/4089

PR #4089, opened 2025-10-30
[Feat]: Add filter `<slot/>` to `p-select`

kaiszybiak:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/3988/

#### Scope

#### Resolved Issue

Resolves [#3988]

#### Questions
- Automatically reset filter on close or emit toggle event to let the consumer decide? => toggle event
- Add validation to only allow p-input-search?
- How to allow complex options in p-multi-select? Currently only textContent joined with `,` is rendered

### DIFF
--- a/packages/components-angular/package.json
+++ b/packages/components-angular/package.json
@@ -41,7 +41,7 @@
     "@angular/platform-browser": "^20.3.4",
     "@angular/platform-browser-dynamic": "^20.3.4",
     "@angular/router": "^20.3.4",
-    "@porsche-design-system/components-angular": "3.31.0-rc.0",
+    "@porsche-design-system/components-angular": "3.31.0",
     "rxjs": "^7.8.2",
     "tslib": "^2.8.1",
     "zone.js": "~0.15.1"
--- a/packages/components-angular/projects/angular-wrapper/package.json
+++ b/packages/components-angular/projects/angular-wrapper/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@porsche-design-system/components-angular",
-  "version": "3.31.0-rc.0",
+  "version": "3.31.0",
   "description": "Porsche Design System is a component library designed to help developers create the best experience for software or services distributed by Dr. Ing. h.c. F. Porsche AG.",
   "keywords": [
     "porsche",
@@ -17,7 +17,7 @@
   "license": "SEE LICENSE IN LICENSE",
   "homepage": "https://designsystem.porsche.com",
   "dependencies": {
-    "@porsche-design-system/components-js": "3.31.0-rc.0",
+    "@porsche-design-system/components-js": "3.31.0",
     "tslib": "^2.8.1"
   },
   "peerDependencies": {
--- a/packages/components-js/package.json
+++ b/packages/components-js/package.json
@@ -85,8 +85,8 @@
   },
   "dependencies": {
     "@porsche-design-system/assets": "5.0.2",
-    "@porsche-design-system/components": "3.31.0-rc.0",
-    "@porsche-design-system/components-js": "3.31.0-rc.0",
+    "@porsche-design-system/components": "3.31.0",
+    "@porsche-design-system/components-js": "3.31.0",
     "@porsche-design-system/components-manager-cli": "0.0.0",
     "@porsche-design-system/partials": "0.0.0",
     "imask": "^7.6.1"
--- a/packages/components-js/projects/components-wrapper/package.json
+++ b/packages/components-js/projects/components-wrapper/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@porsche-design-system/components-js",
-  "version": "3.31.0-rc.0",
+  "version": "3.31.0",
   "description": "Porsche Design System is a component library designed to help developers create the best experience for software or services distributed by Dr. Ing. h.c. F. Porsche AG.",
   "keywords": [
     "porsche",
--- a/packages/components-react/package.json
+++ b/packages/components-react/package.json
@@ -28,7 +28,7 @@
     "clean:build": "rm -rf ./dist && yarn --cwd projects/nextjs clean:build && yarn --cwd projects/remix clean:build"
   },
   "dependencies": {
-    "@porsche-design-system/components-react": "3.31.0-rc.0",
+    "@porsche-design-system/components-react": "3.31.0",
     "react-imask": "^7.6.1",
     "react-router-dom": "^7.9.4",
     "styled-components": "^6.1.19",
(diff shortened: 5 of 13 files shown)


packages/components-vue/src/examples/SelectExampleAsyncFilter.vue
```diff
@@ -0,0 +1,77 @@
...
+    loading.value = false;
+    const newOptions = Array.from({ length: 3 }, (_, i) => ({
+      value: `${term}-${i + 1}`,
+      label: `Result ${i + 1} for "${term}"`,
+    }));
+    setOptions(newOptions);
+  };
+
+  const onInput = (event: InputSearchInputEventDetail) => {
+    const term = (event.target as HTMLInputElement).value;
```
henrijoss: Use `e.target.value`

packages/components-vue/src/examples/SelectExampleAsyncFilter.vue
```diff
@@ -0,0 +1,77 @@
...
+    lastSubmittedData.value = formData.get('options')?.toString() || 'none';
+  };
+</script>
+
+<template>
+  <p-select name="options" label="Async Search">
+    <PInputSearch
+      slot="filter"
+      name="search"
+      v-model="searchValue"
```
henrijoss: v-model is only supported with prop as argument 

```suggestion
      v-model:value="searchValue"
```

packages/components-react/src/examples/SelectExampleAsyncLoad.tsx
```diff
@@ -0,0 +1,57 @@
...
+    }
+  };
+
+  return (
+    <PSelect
+      name="async-select"
+      label="Async Load on Open"
+      value={selectedValue}
+      onChange={onChange}
+      onToggle={(e) => onToggle(e as CustomEvent<SelectToggleEventDetail>)}
```
henrijoss: Maybe onClick would be enough instead of having a new event?

packages/components/src/components/select/select/select.tsx
```diff
@@ -582,11 +606,12 @@ export class Select {
...
   private onToggle = (): void => {
-    if (this.isOpen && this.filter) {
-      // Double requestAnimationFrame as Safari fix to make sure the input will receive focus
+    this.toggle.emit({ value: this.isOpen });
+    if (this.isOpen && (this.filter || this.filterSlot)) {
+      // Double requestAnimationFrame as a Safari fix to make sure the input will receive focus
       requestAnimationFrame(() => {
         requestAnimationFrame(() => {
-          this.inputSearchElement.focus();
+          this.inputSearchInputElement.focus();
```
henrijoss: Make sure the button is focused after closing the dropdown again also when the slotted input is used. 

See in the `onComboKeyDown` function:

```
      case 'Close': {
        event.preventDefault();
        this.updateMenuState(false);
        if (this.filter) {
          this.buttonElement.focus();
        }
        break;
      }
```

packages/components/src/components/select/select/select.tsx
```diff
@@ -243,9 +249,14 @@ export class Select {
   }
 
   public componentDidLoad(): void {
-    getShadowRootHTMLElement(this.host, 'slot').addEventListener('slotchange', this.onSlotchange);
-    if (this.filter) {
-      this.inputSearchInputElement = this.inputSearchElement.shadowRoot.querySelector('input');
+    if (this.filterSlot) {
+      const filter = this.filterSlot.assignedElements()[0];
+      filter.addEventListener('keydown', this.onComboKeyDown);
```
henrijoss: Maybe we can simplify the keydown listener by using `@Listen` on the host to listen for any keydown in the component

packages/components-vue/src/examples/SelectExampleAsyncFilter.vue
kaiszybiak: `Vue: Property value does not exist on type EventTarget`

kaiszybiak: I changed it for React

packages/storefront/src/styles/globals.css
```diff
@@ -1,5 +1,6 @@
 @import 'tailwindcss';
 @import '@porsche-design-system/components-react/tailwindcss';
+@source "../../../../node_modules/@porsche-design-system/shared/examples/*.tsx";
```
kaiszybiak, line 3: Why this?

packages/components/src/components/select/select/select.tsx
```diff
@@ -500,10 +525,13 @@ export class Select {
...
     for (const child of Array.from(this.host.children).filter(
-      (el) => el.tagName !== 'SELECT' && el.slot !== 'label' && el.slot !== 'description' && el.slot !== 'message'
+      (el) =>
+        el.tagName !== 'SELECT' &&
+        el.slot !== 'label' &&
+        el.slot !== 'description' &&
+        el.slot !== 'message' &&
+        el.slot !== 'filter'
     )) {
-      throwIfElementIsNotOfKind(this.host, child as HTMLElement, ['p-select-option', 'p-optgroup']);
```
kaiszybiak, line 505: This is still valid for when the `fliter` slot is unused. So maybe use it conditionally instead ?

packages/components-react/projects/react-ssr-wrapper/tests/unit/specs/__snapshots__/react-ssr-wrapper.spec.tsx.snap
```diff
@@ -19479,6 +19481,14 @@ exports[`should render dsr component for PSelect 1`] = `
   [popover]:not(:popover-open) {
     display: none;
   }
+  ::slotted([slot="filter"]) {
```
kaiszybiak, line 19484: Maybe only add conditionally if `hasCustomFilterSlot` is true

packages/components/src/components/multi-select/multi-select/multi-select.tsx
```diff
@@ -500,10 +531,13 @@ export class MultiSelect {
...
     for (const child of Array.from(this.host.children).filter(
-      (el) => el.tagName !== 'SELECT' && el.slot !== 'label' && el.slot !== 'description' && el.slot !== 'message'
+      (el) =>
+        el.tagName !== 'SELECT' &&
+        el.slot !== 'label' &&
+        el.slot !== 'description' &&
+        el.slot !== 'message' &&
+        el.slot !== 'filter'
     )) {
-      throwIfElementIsNotOfKind(this.host, child as HTMLElement, ['p-multi-select-option', 'p-optgroup']);
```
kaiszybiak, line 505: Can we also use `throwIfElementIsNotOfKind` for the filter slot to ensure only a `p-input-search` is allowed?

PR #4264, opened 2026-03-30
Define min/max-width values for Banner component (Issue/4251)

### LINKED ISSUE #4251: Define min/max-width values for `Banner` component
Scope
---
The `Banner` renders to wide, especially on larger viewports. This leads to decreased reading experience of the provided banner contents.

Out of Scope
---

Notes  
---
- Define a mix width based on `ch` unit (e.g. `max-width: 100ch`)
- Eventually define also a min width value if we decide that the banner defines its width related to the content.

Acceptance criteria
---
- [ ] `Banner` has max-width value based on `ch` unit
- [ ] A CSS variable is provided so that teams can also overwrite the max-width

Subtasks
---
- [ ] Task

o-hook:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/4251/

#### Scope

#### Resolved Issue

Resolves #4251

### DIFF
--- a/packages/components-js/tests/a11y/specs/a11ytree/__snapshots__/p-banner-0.aria.yml
+++ b/packages/components-js/tests/a11y/specs/a11ytree/__snapshots__/p-banner-0.aria.yml
@@ -1,5 +1,5 @@
 - status "Heading":
   - heading "Heading" [level=5]
   - paragraph: Description
-- button "Close notification"
-- status
+  - button "Close banner"
+  - status
--- a/packages/components-js/tests/a11y/specs/a11ytree/__snapshots__/p-banner-1.aria.yml
+++ b/packages/components-js/tests/a11y/specs/a11ytree/__snapshots__/p-banner-1.aria.yml
@@ -1,5 +1,5 @@
-- alert "Heading (state = error)":
-  - heading "Heading (state = error)" [level=5]
+- alert "Heading":
+  - heading "Heading" [level=5]
   - paragraph: Description
-- button "Close notification"
-- status
+  - button "Close banner"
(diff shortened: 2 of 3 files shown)


packages/components/src/components/banner/banner-styles.ts
```diff
@@ -23,9 +21,13 @@ import { getCss } from '../../utils';
 /**
  * @css-variable {"name": "--p-banner-position-top", "description": "Position top of banner", "defaultValue": "56px"}
  * @css-variable {"name": "--p-banner-position-bottom", "description": "Position bottom of banner. Only has an effect below breakpoint 's'.", "defaultValue": "56px"}
+ * @css-variable {"name": "--p-banner-max-w", "description": "Defines the max-width of banner", "defaultValue": "100ch"}
+ * @css-variable {"name": "--p-banner-offset-inline", "description": "Defines the offset left and right from the screen", "defaultValue": "Porsche Grid: gridExtendedOffsetBase"}
  */
 const cssVariableTop = '--p-banner-position-top';
 const cssVariableBottom = '--p-banner-position-bottom';
+const cssVariableMaxWidth = '--p-banner-max-w';
+const cssVariableOffsetInline = '--p-banner-offset-inline';
```
stephanschroeter: Let's have a chat if it makes more sense to use `--p-banner-inset-x` and also use it as `inset-inline` position instead of doing calculations with the width. In addition, we could alias `--p-banner-position-top/bottom` to `--p-banner-top` and `--p-banner-bottom` or `--p-banner-inset-bs` and `--p-banner-inset-be`

PR #4279, opened 2026-04-07
[BUG] - Select not updating children inside of optgroup (#4278) - v3

mihaelsmrcek:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/issue/4278

#### Resolved Issue

Resolves #4278


packages/components/src/components/optgroup/optgroup.tsx
```diff
@@ -29,6 +28,7 @@ export class Optgroup {
   @Watch('disabled')
   public handleDisabledChange(): void {
     updateOptionsDisabled(this.host, this.disabled);
+    this.dispatchInternalOptgroupUpdate();
```
henrijoss: Is this necessary to be called when the disabled state changes or is it already handled with the `updateOptionsDisabled`?

packages/components-js/tests/e2e/specs/multi-select.e2e.ts
```diff
@@ -2446,6 +2462,138 @@ test.describe('optgroups', () => {
...
+
+  test('should select option inside optgroup by click', async ({ page }) => {
+    await initMultiSelect(page, { options: { includeOptgroups: true } });
+
+    const host = getHost(page);
+    const buttonElement = getButton(page);
+    const options = getMultiSelectOptions(page);
+
+    await buttonElement.click();
+    await waitForStencilLifecycle(page);
```
henrijoss, line 2473: Should not be necessary when using web first assertions

packages/components-js/tests/e2e/specs/multi-select.e2e.ts
```diff
@@ -2446,6 +2462,138 @@ test.describe('optgroups', () => {
...
+
+    const host = getHost(page);
+    const buttonElement = getButton(page);
+    const optgroups = getMultiSelectOptgroups(page);
+
+    await expect(host).toHaveJSProperty('value', ['a']);
+    await expect(buttonElement.locator('span').first()).toHaveText('Option A');
+
+    // Simulate dynamic option addition (like React re-rendering with an extra option)
+    await addOptionToOptgroup(page, '0', 'd', 'Option D');
```
henrijoss, line 2572: Should this test that the newly added option does not interfere with the old option selection? Since the newly added option is not used anywhere?

packages/components-react/src/examples/MultiSelectExampleDynamicOptgroup.tsx
```diff
@@ -0,0 +1,27 @@
+import { PMultiSelect, PMultiSelectOption, POptgroup } from '@porsche-design-system/components-react';
+import { useEffect, useState } from 'react';
+
+export const MultiSelectExampleDynamicOptgroupPage = (): JSX.Element => {
+  const [value, setValue] = useState<string[]>(['a']);
+  const [options, setOptions] = useState<string[]>([]);
+
+  useEffect(() => {
```
henrijoss: Maybe would be better to just have a button which adds one new option within the optgroup would be easier to understand and also be a minimal reproduction example

packages/components-react/tests/e2e/specs/multi-select.e2e.ts
```diff
@@ -42,3 +48,37 @@ test.describe('form', () => {
...
   });
 });
+
+test.describe('optgroups', () => {
+  test('should correctly reflect value when options inside optgroup change dynamically', async ({ page }) => {
+    await goto(page, 'multi-select-example-dynamic-optgroup');
+    await waitForComponentsReady(page);
+
+    // Let the interval run for a few ticks so options appear
+    await sleep(1500);
```
henrijoss: Instead of having all this logic within the test, just dynamically add an option, select it and check if the select was updated correctly

packages/components/src/components/optgroup/optgroup.tsx
```diff
@@ -57,9 +56,17 @@ export class Optgroup {
           <span id={labelId} role="presentation">
             {this.label}
           </span>
-          <slot />
+          <slot onSlotchange={() => this.dispatchInternalOptgroupUpdate()} />
```
henrijoss: Does this trigger initially? Maybe check the lifecycles with an E2E test

packages/components/CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0),
 
 ## [Unreleased]
 
+- `OptGroup`, `Multi Select`, `Select`: added internal event to detect slots in `OptGroup`
```
henrijoss:
```suggestion
### Fixed

- `Multi Select`, `Select`: not correctly updating when options within `p-optgroup` change dynamically
```

packages/components-js/tests/e2e/specs/multi-select.e2e.ts
mihaelsmrcek: We use this pattern in other parts of this file

henrijoss (approved):

PR #4443, opened 2026-05-19
Automate storefront documentation via `tokens-meta` sub-package

### LINKED ISSUE #4380: Docs for styling solutions (Tailwind, SCSS, Emotion, Vanilla-Extract) at code level via jsdocs and `styles-meta` sub package
### Scope

-

### Out of Scope

_No response_

### Notes

_No response_

### Acceptance criteria

- [ ] Criteria

kauricre:
### Pull Request Checklist

#### References

- Preview: https://designsystem.porsche.com/pr-4380

#### Scope

- Annotated all storefront/tokens component with jscods standard.
- Created a script which reads all storefront/tokens component and extracts needed information (title, description, value...) which were previously hardcoded into `mdx` files into one file containing multiple objects based on which "token" it belongs to (blur, breakpoint, border...)
- Created sub-package which holds onto that object internally and exposes it into storefront/tokens frontend application
- Updated all `mdx` files with new `<TokenTable>` component which takes token object and renders previously hardcoded table in the same way that it was before (design wise).

#### Resolved Issue

Resolves #4380 (partly)


packages/tokens/projects/tokens-meta/src/types/token-meta.ts
henrijoss, line 1: Instead of using category and group it is probably better to structure the tokens meta object similar like we did it with the folders like this:

```
const tokens = {
  color: {
    light: {
      background: {
        name: 'colorBackdropLight',
        value: 'hsl(240 5.3% 14.9% / 0.5)',
        description: '...' 
      }
    }
  }
}
```

packages/tokens/projects/tokens-meta/package.json
henrijoss, line 1: We should add a unit test here which creates a snapshot of the tokens meta

packages/tokens/projects/tokens-meta/scripts/generateTokensMeta.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const grouped: Record<
+  string,
+  { name: string; value: string; description: string; category: string; group?: string }[]
+> = {};
+
+for (const file of tokenFiles) {
+  const content = fs.readFileSync(file, 'utf-8');
+  const nameMatch = content.match(/export const (\w+)/);
+  const descriptionMatch = content.match(/\/\*\*\s*(.+?)\s*\*\//);
```
henrijoss: Instead of using regex to extract the contents we should be able to use TS Apis which would make the whole system more reliable

ts.createSourceFile()
ts.isVariableStatement()
ts.getJSDocCommentsAndTags()

packages/storefront/src/components/common/TokensTable.tsx
```diff
@@ -0,0 +1,95 @@
...
+  group?: string;
+  sort?: SortDirection;
+};
+
+// Convert camelCase token name to kebab-case CSS var, e.g. colorCanvas → --p-color-canvas
+const toCssVar = (name: string): string => `--p-${name.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase()}`;
+
+const isColorToken = (meta: TokenMeta[]): boolean => meta.length > 0 && meta[0].category === 'color';
+
+export const TokensTable = ({ meta, group, sort = 'none' }: TokensTableProps) => {
```
henrijoss: Probably better to just pass an object of TokenMeta objects and render them instead of having group and sort logic here. Should be doable by structuring the tokens meta object hierarchically

package.json
```diff
@@ -63,11 +64,12 @@
...
     "build-prod": "npm run build:core-dependencies-prod && concurrently -n 'components' 'npm run build:components-prod' && concurrently -n 'components-js' 'npm run build:components-js-prod' && concurrently --kill-others-on-fail -n 'components-angular,components-react,components-vue' 'npm run build:components-angular' 'npm run build:components-react' 'npm run build:components-vue' && npm run build:storefront",
-    "build:core-dependencies": "npm run build:shared && npm run build:tokens && npm run build:assets && npm run build:styles && npm run build:utilities && npm run build:component-meta",
-    "build:core-dependencies-prod": "npm run build:shared && npm run build:tokens && npm run build:assets && npm run build:styles-prod && npm run build:utilities && npm run build:component-meta",
+    "build:core-dependencies": "npm run build:shared && npm run build:tokens && npm run build:tokens-meta && npm run build:assets && npm run build:styles && npm run build:utilities && npm run build:component-meta",
```
henrijoss: Instead of adding the tokens-meta build here i think it would be better to make it part of the tokens build in `npm run build:tokens`

packages/tokens/projects/tokens-meta/scripts/generateTokensMeta.ts
```diff
@@ -0,0 +1,123 @@
...
+
+const files = await fg(`${sourceDirectory}/**/*.ts`);
+// Exclude barrel files, tests, light/dark theme-specific files (only light-dark variants are used),
+// and the raw palette constants which are not design tokens themselves (as per previous implementation).
+const tokenFiles = files.filter(
+  (f) =>
+    !f.endsWith('index.ts') &&
+    !f.endsWith('.spec.ts') &&
+    !f.includes('/color/light/') &&
+    !f.includes('/color/dark/') &&
```
henrijoss: Why exclude these?

packages/tokens/projects/tokens-meta/scripts/generateTokensMeta.ts
```diff
@@ -0,0 +1,123 @@
...
+    !f.endsWith('palette.ts')
+);
+
+function extractTokenInfo(filePath: string): { name: string; description: string } | null {
+  // setParentNodes=true is required so getJSDocCommentsAndTags() can walk up
```
henrijoss: Comment does not relate to any actual code or is confusing?

packages/tokens/projects/tokens-meta/scripts/generateTokensMeta.ts
kauricre: My original thinking was that we don't need those in the final build since we don't display these in the storefront as separate color modes (and rather unified as `light-dark`). Now I see that it was wrong and they are still legitimate objects and values that should be exported as well. Please correct me if I am wrong.

packages/storefront/src/components/common/TokensTable.tsx
```diff
@@ -0,0 +1,138 @@
...
+const toSortNum = (value: string): number => {
+  const clampMatch = value.match(/clamp\(\s*([\d.]+)/);
+  if (clampMatch) return parseFloat(clampMatch[1]);
+  // shadows: use the largest px value as the size indicator
+  const pxValues = Array.from(value.matchAll(/([\d.]+)px/g), (m) => parseFloat(m[1]));
+  if (pxValues.length > 0) return Math.max(...pxValues);
+  return parseFloat(value);
+};
+
+export const TokensTable = ({ meta, sort }: TokensTableProps) => {
```
henrijoss: This seems to be more complex than necessary. We dont need to sort or extract anything. We already have the color value which can directly be used. Add a property to specify that a swatch for colors should be displayed like `showColorSwatch`.

packages/tokens/projects/tokens-meta/scripts/generateTokensMeta.ts
```diff
@@ -0,0 +1,123 @@
...
+  let node = tree;
+  for (const seg of segments) {
+    if (!node[seg]) node[seg] = {};
+    node = node[seg] as TokenTree;
+  }
+
+  node[info.name] = {
+    name: info.name,
+    // String() handles numeric token values (e.g. breakpoints) that aren't already strings
+    value: String(resolvedValue),
```
henrijoss: We should be able to keep the actual values instead of stringifying them

packages/tokens/projects/tokens-meta/src/types/token-meta.ts
```diff
@@ -0,0 +1,5 @@
+export type TokenMeta = {
+  name: string;
+  value: string;
```
henrijoss: value can be of whatever types we have in our tokens e.g. string | number etc

package.json
```diff
@@ -67,7 +68,8 @@
     "build:core-dependencies-prod": "npm run build:shared && npm run build:tokens && npm run build:assets && npm run build:styles-prod && npm run build:utilities && npm run build:component-meta",
     "build:assets": "npm run build --workspace=@porsche-design-system/assets",
     "build:assetsCopyComponentsOnly": "npm run buildCopyComponentsOnly --workspace=@porsche-design-system/assets",
-    "build:tokens": "npm run build --workspace=@porsche-design-system/tokens",
+    "build:tokens": "npm run build --workspace=@porsche-design-system/tokens && npm run build:tokens-meta",
```
henrijoss: The tokens meta should be build when running `npm run build --workspace=@porsche-design-system/tokens`, so within the package.json of the token package, since we always want to generate the meta whenever we rebuild the tokens package

packages/storefront/src/components/common/TokensTable.tsx
```diff
@@ -0,0 +1,72 @@
...
+        </PTableRow>
+      </PTableHead>
+      <PTableBody>
+        {items.map((token) => (
+          <PTableRow key={token.name}>
+            {showColorSwatch && (
+              <PTableCell>
+                <div
+                  className="w-10 h-10 rounded-md border border-contrast-low"
+                  style={{ background: `var(${toCssVar(token.name)})` }}
```
henrijoss: Probably better to just use the token value directly since its not guaranteed that the token name will match the css var from our global styles

henrijoss (approved):

PR #4458, opened 2026-05-24
feat(components-js): add exclude option to load()

### LINKED ISSUE #4457: p-canvas main slot: support for flex-based height-constrained layouts
Hi PDS team, greetings from FDP4!

First off: great work on v4, the new canvas layout and the overall DX improvements are really solid.

We have one request regarding `p-canvas`. We are building data tables (PrimeNG) that need to fill the available viewport height with a scrollable body and a fixed paginator at the bottom. This requires an unbroken flex/height chain from the viewport down to the scroll container.

The `<main class="main">` element inside the `p-canvas` Shadow DOM currently uses `overflow: visible` and `display: block`, which breaks that chain. Slotted content grows to its intrinsic height instead of being constrained by the grid row.

Would it be possible to either:

1. Change the default styles on `<main>` to support flex-based layouts (e.g. `display: flex; flex-direction: column; overflow: hidden` with `::slotted(*) { flex: 1 1 auto; min-height: 0 }`)
2. Or expose an attribute/property on `p-canvas` to opt into this behavior

We currently work around this by patching the Shadow DOM after `componentsReady()`, which works but feels fragile.

PDS version: 4.1.0

Thanks!

flgke81:
### Pull Request Checklist

#### References

- Related issue: #4457

#### Scope

Expose the existing `exclude` array from Stencil's `bootstrapLazy()` through the public `load()` API. Consumers can prevent PDS from registering specific custom elements that collide with third-party libraries (PrimeNG, Ionic, Shoelace).

PDS uses the `p-` prefix. PrimeNG uses identical selectors (`p-table`, `p-select`, `p-checkbox`, etc.) that are hardcoded and cannot be changed. When PDS registers these tags first, it wraps them in Shadow DOM and breaks PrimeNG's light-DOM rendering. 16 tag names collide today.

Stencil already has the mechanism internally:

```javascript
// bootstrapLazy() in the generated bundle
const exclude = options.exclude || [];
if (!exclude.includes(tagName) && !customElements2.get(tagName)) {
  customElements2.define(tagName, proxyComponent(...));
}
```

This PR pipes `exclude` from `load()` through `loadComponentLibrary()` and the CDN script entry point into `defineCustomElements()`, where it reaches `bootstrapLazy()`.

Consumer usage after this change:

```typescript
load({ exclude: ['p-table', 'p-scroller', 'p-select', 'p-checkbox'] });
```

Changes in 3 files:

- `components-manager-core/src/services/library-handler.ts` - add `exclude: string[]` to types, store and merge in version data
- `components-manager-cli/library-entry/with-prefix.ts` - add `exclude?: string[]` to `LoadOptions`, pass through
- `components-wrapper/src/index.js` - read `exclude` from version data in the `setRegisterComponentsCallback`, pass to `defineCustomElements()`

#### Resolved Issue

Resolves #4457

### DIFF
--- a/packages/components-js/projects/components-manager-cli/library-entry/with-prefix.ts
+++ b/packages/components-js/projects/components-manager-cli/library-entry/with-prefix.ts
@@ -31,10 +31,12 @@ declare var CM_CONFIG: EntryConfig;
 /**
  * @property prefix - the prefix used for the components
  * @property cdn - the cdn to load assets from
+ * @property exclude - tag names to skip during custom element registration (e.g. ['p-table', 'p-select'])
  */
 export type LoadOptions = {
   prefix?: string;
   cdn?: 'auto' | 'cn';
+  exclude?: string[];
 };
 
 export const load = (opts: LoadOptions = {}): void => {
@@ -58,5 +60,5 @@ export const load = (opts: LoadOptions = {}): void => {
     prefixes: [], // to not break older versions
   };
 
-  loadComponentLibrary({ ...CM_CONFIG, prefix: opts.prefix || '' });
+  loadComponentLibrary({ ...CM_CONFIG, prefix: opts.prefix || '', exclude: opts.exclude });
 };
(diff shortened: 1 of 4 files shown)


henrijoss:
You can specify a custom prefix in the provider to avoid any collisions:
https://designsystem.porsche.com/v4/developing/react/advanced/#prefixing

`<PorscheDesignSystemProvider prefix="sample-prefix">`

Issue #4107, opened 2025-11-12
Enhance PDS docs to show a11y do's (best practices) and dont's


o-hook:
Scope
---
Enhance the current docs with best practices and do's/dont's also on component level, e.g.:

**General Guides:**
- Accessible Form Validation (https://blog.pope.tech/2025/09/30/accessible-form-validation-with-examples-and-code/)
- Structuring your code with ARIA landmarks (section, article, nav, aside) and name them
- Add skip to content link or content skip links
- how to hide contents visually but still accessible for screen readers (tailwind calls and other solutions: https://www.a11y-collective.com/blog/visually-hidden/)

**Component specific:**
- Don't ad ARIA attributes on component root level (use aria prop instead)
- Add mandatory aria props (e.g. on a button if a modal is opened)
- How to use skip link in carousels

Out of Scope
---
Do not document how to use HTML in deep (link to further guidelines)

Notes  
---
When conducting A11Y Audits, the same mistakes are found again and again. So it might sense to give more guidelines and info on how to do it the right way.

Acceptance criteria
---
- [ ] Section for general guidelines are integrated into Storefront
- [ ] Additional guidelines and hints are integrated on component level

Subtasks
---
- [ ] Task

PR #4562, opened 2026-07-02
issue/4516

### LINKED ISSUE #4516: Turn Popover into controlled component
### Scope

Team navigate want so inform users about the market version ob a webpage that is current displayed. Therefore they plan to use sort of a tooltip with functionalities which the tooltip component of the PDS does not support. See Example below.

Requested Extentsions:
- The popover must be dismissible. A dismiss button should be implemented to allow users to close the popover
- Stays visible when user interacting with content outside of tool tip
- Can be displayed without requiring a hover action (e.g. on page load immediately visible)
- Will automatically disappear after (configurable) seconds and/or when the user scrolls down (to be discussed)
- Popover can be wider to display desired content correctly
- Enable canvas and surface as bg color for popover (?)
- Toggle drop shadow

- The popover should accommodate longer descriptions (already supported)
- A pointer can be added or removed and can be centered, left-aligned, or right-aligned (use current logic implemented)
- Must support a call to action (already supported)

### Out of Scope

_No response_

### Notes

We need to make sure that the current logic of the popover is kept, and requested behaviors of component can be toggled e.g. immediately visible, time base hiding, persistent display while user interaction with other content…

### Acceptance criteria

- [ ] Consumer can define and control behavior of popover (visible on hover, immediately, time based,…)
- [ ] Current display logic is maintained (show on click, hide on interaction out side of popover)
- [ ] Consumer can place a dismiss button inside of the tool tip to close it
- [ ] Consumer can define size of the popover 

Image

Image

stephanschroeter:
### Pull Request Checklist

Preview: https://designsystem.porsche.com/pr-4562/components/popover/configurator/

### Related Issue

Closes #4516


packages/components-js/src/pages/popover-compact.html
```diff
@@ -0,0 +1,20 @@
...
+    gap: 64px;
+  }
+</style>
+
+<p-popover open="true" compact="true" direction="right">
+  <p-text>Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat.</p-text>
+</p-popover>
+
+<p-popover open="true" compact="true" direction="left">
+  <p-button-pure slot="button" hide-label="true" icon="car" size="small">More information</p-button-pure>
```
o-hook: add `aria="{'aria-expanded': 'true'}"` to the trigger button

o-hook (commented):
A11Y specific docs will be updated with #4107

PR #4572, opened 2026-07-07
Migrate away from unmaintaned `JSS` library

### LINKED ISSUE #4532: Migrate away from JSS (unmaintained) to unblock `@babel/runtime` v8 and future dependency updates
## Summary

`jss` (and its plugins + `css-vendor`) is unmaintained and is now blocking dependency upgrades — most immediately `@babel/runtime@8`. We currently pin `@babel/runtime` to v7 via a root override to keep things working, but this is a stop-gap. Since JSS needs to be replaced anyway, we should plan that migration and remove the v7 pin afterward.

## Background

- Vite 8 (rolldown) strictly enforces package `exports`. `@babel/runtime@8` **removed the `./helpers/esm/*` export subpaths**.
- `jss@10.10.0`, several `jss-plugin-*`, and `css-vendor` were compiled against Babel 7 and hard-import those removed subpaths (e.g. `helpers/esm/extends`, `helpers/esm/objectWithoutPropertiesLoose`, `helpers/esm/toConsumableArray`).
- A global `"@babel/runtime": "^8.0.0"` override forced v8 onto these packages and broke `build-app` (and would affect the wider build):

  ```
  "./helpers/esm/extends" is not exported … from jss-plugin-nested/node_modules/@babel/runtime
  ```

- These JSS packages are effectively dormant (no Babel 8 / updated builds available).

## Current workaround (already in place)

- Root override set to `"@babel/runtime": "^7.29.7"` (v7 still exposes both `helpers/*` and `helpers/esm/*` and satisfies every dependent's `^7.x` range).
- This is intentionally not the long-term solution.

## Why migrate (not just keep pinning)

- JSS is unmaintained → growing compatibility risk with the toolchain (Vite/rolldown, Babel, Node).
- Keeping the v7 pin blocks adopting `@babel/runtime@8` globally and adds maintenance overhead (scoped overrides/patches).
- Aligns with our goal of keeping dependencies current.

## Scope / where JSS is used

- Core component styling in `packages/components` (the `*-styles.ts` files use JSS at runtime to build stylesheets).
- Transitive: `jss`, `jss-preset-default`, `jss-plugin-*`, `css-vendor`.

## Proposed direction

1. Decide on a replacement styling approach consistent with the existing stack (e.g. the Tailwind / `vanilla-extract` / `emotion` tooling already present in `packages/styles`), and define migration guidelines.
2. Migrate component styles off JSS incrementally, preserving accessibility behavior (focus styles via `getFocusBaseStyles()`, forced-colors/HCM support) and VRT parity.
3. Remove `jss*` + `css-vendor` from the dependency tree.
4. Remove the `@babel/runtime` v7 override and verify the build with v8.

## Interim options (until migration lands)

- **Keep** the `^7.29.7` override (current, simplest), **or**
- Scope a v7 override only to the JSS ecosystem and let everything else move to v8, **or**
- `patch-package` `jss`/`css-vendor` to import non-esm helpers.

## Acceptance criteria

- [ ] Replacement styling solution chosen and documented.
- [ ] All component styles migrated off JSS (no `jss*`/`css-vendor` in `node_modules`).
- [ ] Accessibility preserved (focus, forced-colors/HCM) and VRT/a11y tests pass.
- [ ] `@babel/runtime` v7 override removed; build green on v8.
- [ ] No bundle-size or runtime regressions in `components`/`components-js`.

## References

- Root override: `package.json` → `overrides["@babel/runtime"]`
- Failing build step: `packages/components-js` → `npm run build-app` (`vite build`)

### Out of Scope

_No response_

### Notes

_No response_

### Acceptance criteria

- [ ] Criteria

kauricre:
### Pull Request Checklist

### Related Issue

Closes #4532


packages/components/src/utils/css-serializer.ts
```diff
@@ -0,0 +1,117 @@
...
+//
+// Don't swap this for emotion or another CSS-in-JS lib: it runs on the same stylis, so it
+// wouldn't hoist/combine media queries either (we'd still need hoistAndCombineMedia below),
+// and it adds class hashing/caching we don't use. More weight, no gain.
+import { compile, serialize, stringify } from 'stylis';
+
+// Loose stand-ins for jss's `Styles`/`JssStyle` types. Kept as `any` on purpose so the
+// ~50 existing *-styles.ts files don't need to be retyped against strict CSS types.
+export type JssStyle = Record<string, any>;
+export type Styles<_Name = string, _Data = unknown, _Theme = undefined> = Record<string, any>;
```
stephanschroeter: Are equivalent types from Stylis available?

kauricre: No since `stylis` operates on CSS strings, not style objects. There is no proper type from `@types/stylis` package either.

PR #4686, opened 2026-08-24
Fix `p-carousel` rendering issues

### LINKED ISSUE #4680: p-carousel drops slides when the slide count changes without changing the page count
### JS Framework

Vue 3 (inside Astro) — but the defect is framework-agnostic, it lives in the `p-carousel` component source.

4.5.0 (also present on `main` at the time of writing)

Mac (Sonoma), any desktop/tablet viewport. Reproducible on any device — it is viewport-width dependent, not device dependent.

Chrome (verified with `HeadlessChrome/151.0.0.0` and desktop Chrome). Browser-agnostic — the cause is a missing Stencil re-render, not a browser quirk.

### Browser version

151.0.0.0

### Reproduction Example

No StackBlitz — the bug is reproducible on a public production page and, more importantly, it is directly identifiable in the component source (see "Root cause" below).

**Live reproduction:** https://www.porsche.com/germany/models/911/carrera-models/911-carrera/

**Minimal conceptual reproduction:**

```html
<p-carousel slides-per-page='{"base": 1, "s": 2, "l": 3}'>
  <div>Slide A</div>
</p-carousel>
```

At a viewport where `slidesPerPage` resolves to `3`, replace the single child with two children:

```js
const carousel = document.querySelector('p-carousel');
carousel.replaceChildren(
  Object.assign(document.createElement('div'), { textContent: 'Slide A' }),
  Object.assign(document.createElement('div'), { textContent: 'Slide B' })
);

await new Promise((r) => setTimeout(r, 500));

carousel.shadowRoot.querySelectorAll('.splide__slide slot').length; // => 1 (expected: 2)
carousel.children[1].assignedSlot;                                  // => null  ❌ never rendered
```

The second child ends up with `slot="slide-1"` in the light DOM, but no `<slot name="slide-1">` exists in the shadow DOM, so it is silently never projected.

1. Open https://www.porsche.com/germany/models/911/carrera-models/911-carrera/ at a viewport width of at least 1300px (so `slidesPerPage` resolves to `3`).
2. Scroll down to the "Soft-Compare" module and wait for it to hydrate. It renders a `p-carousel` whose slides are swapped by a `p-tabs-bar` above it.
3. Select the body type **GT Cabriolet** (renders **1** slide).
4. Select the body type **Targa** (should render **2** slides).
5. Only **1** card is visible. The second card ("911 Targa 4 GTS") is present in the light DOM with `slot="slide-1"`, but `assignedSlot === null`.

The reverse case also fails: go **GT** (4 slides) → **Targa** (2 slides) → **Turbo Coupé** (1 slide) and you get a surplus empty slide plus an extra pagination bullet.

### Expected behavior

`p-carousel` should always render exactly one `<slot name="slide-i">` per slotted light-DOM child, regardless of whether the resulting page count changes.

---

## Root cause

`render()` builds the shadow-DOM slide wrappers from `this.slides`:

```tsx
{this.slides.map((_, i) => (
  <div key={i} class="splide__slide" tabIndex={0}>
    <slot name={`slide-${i}`} />
  </div>
))}
```

But `slides` is **not** a reactive `@State()` — only `amountOfPages` is:

```ts
@State() private amountOfPages: number;
// ...
private slides: HTMLElement[] = [];   // ← not reactive
```

The `MutationObserver` on the host calls:

```ts
private updateSlidesAndPagination = (): void => {
  this.slides = getSlidesAndAddAttributes(this.host); // re-assigns slot="slide-n" on children
  this.updateAmountOfPages();                        // only this can trigger a re-render
};
```

So a re-render only happens when `amountOfPages` changes value. And `getAmountOfPages()` is not injective for small slide counts:

```ts
export const getAmountOfPages = (amountOfSlides: number, slidesPerPage: number): number =>
  amountOfSlides === 0 ? 0 : amountOfSlides < slidesPerPage ? 1 : amountOfSlides - slidesPerPage + 1;
```

For **any** two slide counts `n1, n2` with `0 < n1, n2 <= slidesPerPage`, it returns `1` for both → no state change → **no re-render** → the number of `<slot>` elements stays stale.

- `n2 > n1` → surplus children get slot names with no matching shadow slot → **slides silently disappear**.
- `n2 < n1` → surplus `<slot>`s remain → **empty slide + extra pagination bullet**.

`splide.refresh()` is still called from `updateAmountOfPages()`, but it cannot help: the slide wrapper elements themselves were never re-rendered.

### Isolating proof

On the broken instance, forcing an *unrelated* Stencil re-render — without touching a single child — immediately heals it:

```js
carousel.gradient = !carousel.gradient;
// before: { childrenInLightDom: 2, slotsInShadowDom: 1, actuallyRendered: 1 }
// after:  { childrenInLightDom: 2, slotsInShadowDom: 2, actuallyRendered: 2 }  ✅
```

This confirms the children and their `slot` attributes were always correct; only the re-render was missing.

### Viewport dependency

With `slidesPerPage = { base: 1, s: 2, l: 3 }`:

| Breakpoint | slidesPerPage | Affected slide-count transitions |
| --- | --- | --- |
| `base` | 1 | none (`getAmountOfPages(n, 1) === n` is injective) |
| `s` | 2 | 1 ↔ 2 |
| `l` | 3 | any pair within {1, 2, 3} |

### Suggested fix

Track the slide *count* as reactive state, so `render()` re-runs exactly when the slot list needs to change:

```diff
   @State() private amountOfPages: number;
+  @State() private slideCount = 0;

   private splide: Splide;
   private slides: HTMLElement[] = [];
```

```diff
   private updateSlidesAndPagination = (): void => {
     this.slides = getSlidesAndAddAttributes(this.host);
+    this.slideCount = this.slides.length;
     this.updateAmountOfPages();
   };
```

```diff
-  {this.slides.map((_, i) => (
+  {Array.from({ length: this.slideCount }, (_, i) => (
     <div key={i} class="splide__slide" tabIndex={0}>
       <slot name={`slide-${i}`} />
     </div>
   ))}
```

A separate `slideCount` is proposed rather than decorating `slides` itself with `@State()`, because `getSlidesAndAddAttributes()` returns a new array instance on every call and the observer also watches `subtree` and `characterData` — making `slides` reactive would cause a re-render on every observed mutation. `componentDidUpdate()` already calls `splide.refresh()`, so no further change is needed there.

### Impact

Any consumer that swaps `p-carousel` slides dynamically (filters, tabs, category switches) can silently hide content. In our case, whole car models disappear from a comparison module on porsche.com.

### Relevant log output

```shell
// Measured on https://www.porsche.com/germany/models/911/carrera-models/911-carrera/
// viewport 1600x1000  ->  slidesPerPage = 3

[
  { tab: 'Turbo Cabriolet', childrenInLightDom: 1, slotsInShadowDom: 1, actuallyRendered: 1 },
  { tab: 'Targa',           childrenInLightDom: 2, slotsInShadowDom: 1, actuallyRendered: 1 },  // <- 1 slide hidden
  { tab: 'GT',              childrenInLightDom: 4, slotsInShadowDom: 4, actuallyRendered: 4 },
  { tab: 'Targa',           childrenInLightDom: 2, slotsInShadowDom: 2, actuallyRendered: 2 },
  { tab: 'Turbo Coupe',     childrenInLightDom: 1, slotsInShadowDom: 2, actuallyRendered: 1 }   // <- 1 empty slide
]

// slot assignment in the broken "Targa" state:
[
  { slot: 'slide-0', assignedSlot: true,  content: '911 Targa 4S' },
  { slot: 'slide-1', assignedSlot: false, content: '911 Targa 4 GTS' }   // never projected
]
```

kauricre:
### Pull Request Checklist

### Related Issue

Closes #4680

### DIFF
--- a/packages/components/src/components/carousel/carousel.tsx
+++ b/packages/components/src/components/carousel/carousel.tsx
@@ -139,7 +139,8 @@ export class Carousel {
   /** Overrides the default label strings used for the previous, next, and page indicators — useful for localization. */
   @Prop() public intl?: CarouselInternationalization;
 
-  @Prop({ mutable: true }) public activeSlideIndex?: number = 0;
+  /** Sets the zero-based index of the currently visible slide. Update this to navigate programmatically. */
+  @Prop() public activeSlideIndex?: number = 0;
 
   /** Sets the `href` of an in-page skip link that lets keyboard users jump past the carousel slides. */
   @Prop() public skipLinkTarget?: string;
@@ -412,7 +413,6 @@ export class Carousel {
     });
 
     splide.on('move', (activeIndex, previousIndex): void => {
-      this.activeSlideIndex = activeIndex;
       updatePrevNextButtons(this.btnPrev, this.btnNext, splide);
       updatePagination(this.paginationEl, this.getPageCount(), activeIndex);
       this.update.emit({ activeIndex, previousIndex });
--- a/packages/components/src/components/carousel/carousel.tsx
+++ b/packages/components/src/components/carousel/carousel.tsx
@@ -190,6 +190,17 @@ export class Carousel {
     this.splide.go(newValue); // change event is emitted via splide.on('move')
   }
 
+  @Watch('slidesPerPage')
+  public slidesPerPageHandler(): void {
+    if (this.splide) {
+      // splideJS reads its breakpoints only when constructed, so a new instance is needed to apply the new value
+      const { index } = this.splide;
+      this.splide.destroy();
+      this.initSplide(index); // splideJS clamps the start index if it exceeds the new amount of pages
+    }
+    this.updateAmountOfPages();
+  }
+
   public connectedCallback(): void {
     observeChildren(this.host, () => {
       this.updateSlidesAndPagination();
@@ -221,31 +232,7 @@ export class Carousel {
 
   public componentDidLoad(): void {
     this.observeSlides(); // initial, adjust aria attributes on slides
-    this.splide = new Splide(this.container, {
-      start: this.activeSlideIndex,
-      arrows: false,
-      easing: motionEasingBase,
-      focus: this.focusOnCenterSlide ? 'center' : undefined,
-      trimSpace: this.trimSpace,
-      pagination: false,
-      rewind: this.rewind,
-      rewindByDrag: true, // only works when rewind: true
-      drag: this.hasNavigation,
-      perMove: 1,
-      mediaQuery: 'min',
-      speed: Number.parseFloat(carouselTransitionDuration) * 1000,
-      gap: gridGap,
-      live: false,
-      // TODO: this uses matchMedia internally, since we also use it, there is some redundancy
-      breakpoints: getSplideBreakpoints(
-        this.parsedSlidesPerPage as Exclude<BreakpointCustomizable<CarouselSlidesPerPage> | 'auto', string>
-      ),
-      // https://splidejs.com/guides/i18n/#default-texts
-      i18n: parseJSONAttribute(this.intl || {}), // can only be applied initially atm
-      direction: getLangDirection(this.host),
-    });
-
-    this.registerSplideHandlers(this.splide);
+    this.initSplide(this.activeSlideIndex);
   }
 
   public componentDidUpdate(): void {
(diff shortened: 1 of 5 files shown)


packages/components/src/components/carousel/carousel.tsx
```diff
@@ -178,8 +190,32 @@ export class Carousel {
     this.splide.go(newValue); // change event is emitted via splide.on('move')
   }
 
+  @Watch('slidesPerPage')
+  public slidesPerPageHandler(): void {
+    if (this.splide) {
+      // splideJS reads its breakpoints only when constructed, so a new instance is needed to apply the new value
+      const { index } = this.splide;
```
henrijoss, line 201: index is the currently active slide? `this.activeSlideIndex` seems to be never updated, which predates this change but should be updated and kept in sync?

henrijoss:
CHANGELOG missing

packages/components/src/components/carousel/carousel.tsx
```diff
@@ -178,14 +194,41 @@ export class Carousel {
...
+    );
     this.observeBreakpointChange();
 
     if (this.splide) {
       this.observeSlides(); // on reconnect, adjust aria attributes on slides
       // on reconnect we can reuse the splide instance
       this.updateSlidesAndPagination();
+      // splideJS reads `start` when it mounts, so a destroyed and re-mounted instance would
+      // otherwise jump back to the index it was constructed with instead of where the user is
+      this.splide.options = { start: this.activeSlideIndex };
```
henrijoss: Since `this.activeSplideIndex` is currently not kept in sync with the actual index does this make sense?

henrijoss (approved):

PR #4692, opened 2026-08-25
Patch `Stencil` to accept `cloneNode` for our shadow root

### LINKED ISSUE #4676: Support cloneNode on shadow DOM components for AG-Grid auto-resize compatibility
### Portfolio

Customer

### ART and Team

ART Physical Products / Team Rainbow

### Problem

We started using AG-Grid on our project and are running into breakages caused by PDS components using shadow DOM.

AG-Grid has a mechanism to auto-resize columns on first render based on content size. Internally, it does this by copying ("cloning") the table content into a hidden phantom table to measure width, then applies that info back to the real table. It does this via the native `cloneNode` function.

The problem: `cloneNode` does not clone shadow roots by default, so when AG-Grid tries to clone a PDS component that uses shadow DOM, the clone comes out empty/broken, breaking the auto-resize feature.

We looked for a workaround on our side but couldn't find one that wasn't over-engineered.

**Solution ideas and examples**:
The `ShadowRoot` interface now supports a `clonable` property (see [MDN: ShadowRoot.clonable](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/clonable)). If PDS components set `clonable: true` when attaching their shadow root, `cloneNode(true)` would correctly clone the shadow DOM content, fixing this AG-Grid integration issue (and likely other cloneNode-based tooling too).

### Solution ideas and examples

The `ShadowRoot` interface now supports a `clonable` property (see [MDN: ShadowRoot.clonable](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/clonable)). If PDS components set `clonable: true` when attaching their shadow root, `cloneNode(true)` would correctly clone the shadow DOM content, fixing this AG-Grid integration issue (and likely other cloneNode-based tooling too).

### Business priority

Medium

### Available assisting resources

_No response_

kauricre:
### Pull Request Checklist

### Related Issue

Closes #4676

### DIFF
--- a/packages/components/CHANGELOG.md
+++ b/packages/components/CHANGELOG.md
@@ -14,6 +14,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0), 
 
 ## [Unreleased]
 
+### Added
+
+- **Shadow DOM**: shadow roots are now attached with `clonable: true`, so `cloneNode(true)` on a component returns a
+  clone that carries its Shadow DOM content instead of an empty shadow root.
+  ([#4692](https://github.com/porsche-design-system/porsche-design-system/pull/4692))
+
 ### Fixed
 
 - `Carousel`: slides were dropped or left over when the amount of slides changed without also changing the amount of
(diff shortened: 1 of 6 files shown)


packages/components/CHANGELOG.md
```diff
@@ -23,6 +23,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0),
 - `Carousel`: changing `slides-per-page` at runtime had no effect. Slide widths and the pagination kept the value the
   carousel was initialised with.
   ([#4686](https://github.com/porsche-design-system/porsche-design-system/pull/4686))
+- **Shadow DOM**: shadow roots are now attached with `clonable: true`, so `cloneNode(true)` on a component returns a
```
henrijoss: Isn't this more of a feature and not a fix?

henrijoss (approved):

PR #4701, opened 2026-09-01
Fix boolean attribute shorthand for BreakpointCustomizable props (Issue/4404)

### LINKED ISSUE #4404: Boolean attribute shorthand (e.g. `hide-label`) not working without explicit value
### JS Framework

Vanilla / any (HTML attribute behavior)

V4

any

Chrome

### Browser version

latest

### Reproduction Example

Reproducible in StackBlitz (internal report). Minimal repro: a PDS component with a boolean prop set as `hide-label` vs `hide-label="true"`.

1. Use any PDS component that exposes a boolean prop, e.g. `<p-text-field-wrapper>` with `hide-label`.
2. Set the attribute without a value: `<p-text-field-wrapper hide-label>...</p-text-field-wrapper>`.
3. Compare with `<p-text-field-wrapper hide-label="true">...</p-text-field-wrapper>`.

### Expected behavior

Per standard HTML semantics, the presence of a boolean attribute should be treated as truthy. `hide-label` and `hide-label="true"` should behave identically. This is also stated in the documentation and used to work previously.

### Relevant log output

_No response_

---

**Additional notes**
- Possibly related to Stencil internals or to props that are breakpoint-customizable.
- Affects API ergonomics and matches standard HTML semantics.

henrijoss:
### Pull Request Checklist

### Related Issue

Closes #4404

### What was done

Boolean props typed as `BreakpointCustomizable<boolean>` did not support the HTML boolean attribute shorthand. Writing
`<p-input-text hide-label>` (without a value) was ignored and logged a validation error, while `hide-label="true"`
worked. Reason: `BreakpointCustomizable<T>` is `T | BreakpointValues<T> | string` — the `string` member exists so that
objects can be passed as attributes, e.g. `size="{ base: 'small', l: 'large' }"`. Because of it, Stencil resolves the
prop type to `any` and skips its own string-to-boolean coercion, so the component received the empty string `''`.

**Fix**

The empty string is mapped to `true`, but only for boolean props. Since `parseJSON()` is type-blind, a blanket
coercion would also hit non-boolean props such as `size` or `aspect-ratio`, where `''` is a plain (invalid) value.
The parse and style helpers are therefore split so that the compiler enforces the correct usage:

- `packages/components/src/utils/breakpoint-customizable.ts`: shared private `parse()`, plus
  - `parseJSON(prop: BreakpointCustomizable<string | number>)` — no shorthand handling
  - `parseJSONBoolean(prop: BreakpointCustomizable<boolean>)` — maps `''` to `true`
- `packages/components/src/utils/jss.ts`: shared private `buildStyles()`, plus
  - `buildResponsiveStyles<T extends string | number>()` — 20 call sites unchanged
  - `buildResponsiveBooleanStyles()` — 28 call sites in 12 style files migrated

  Passing a boolean prop to `buildResponsiveStyles()` is now a compile error, so the shorthand cannot be lost by
  accident and no `valueType` argument has to be passed (and possibly passed wrongly) at 48 call sites.
- `validateProps()` picks the variant via `allowedValues === 'boolean'`, so a non-boolean prop keeps reporting its
  actual value instead of `true` in the validation error.
- Added `isTruthyForAnyBreakpoint()` and used it in `getLinkButtonStyles()`, where `hideLabel` was previously
  evaluated as a raw truthy check. That was wrong for both the shorthand (`''` is falsy) and for breakpoint objects
  such as `{ base: false, l: true }` (always truthy). Same fix applied to `throwIfAlignTopAndNotCompact()`.
- Replaced the duplicated `this.compact === 'true' ? … : …` workarounds (and their `TODO` comments) in `p-link-tile`
  and `p-button-tile` with a shared `getParsedTileCompact()` util.

Note on naming: the helper is called `parseJSONBoolean` rather than `parseBooleanJSON` because
`scripts/generateDSRComponents.ts` strips lines matching `parseJSON` when generating the DSR/SSR components. Verified
by regenerating them — the output is byte-identical.

**Follow-up**

This remains a workaround: as long as objects can be passed as attribute strings, `BreakpointCustomizable<T>` has to
include `string`, and an attribute can never distinguish `hide-label` (shorthand) from `hide-label=""` (empty string
value). #4708 tracks the clean long-term solution for v5 — allow objects via property only and drop attribute object
support — which would remove the `string` member and make `parseJSONBoolean()`, `buildResponsiveBooleanStyles()` and
the new meta-test obsolete. Referenced via `TODO: [v5] … see #4708` comments in the affected files.

Stencil's `@AttrDeserialize()` / `@PropSerialize()` ([docs](https://stenciljs.com/docs/serialization)) was evaluated
as an alternative and rejected for now: the decorators only run on attribute → property changes, while all three
framework wrappers set properties, and the DSR/SSR components and the jsdom-polyfill are generated separately and
would not execute them.

**Tests**

- Unit tests for `parseJSON()`, `parseJSONBoolean()`, `isTruthyForAnyBreakpoint()`, `buildResponsiveStyles()`,
  `buildResponsiveBooleanStyles()`, `validateProps()` and `throwIfAlignTopAndNotCompact()` covering the empty-string
  case.
- New meta-test asserting that no `BreakpointCustomizable` type mixes `boolean` with other types, since such a union
  would be unfixable — an attribute cannot distinguish the shorthand from an empty string value.
- New e2e spec `boolean-attribute-shorthand.e2e.ts` plus example pages for JS, Angular and Vue, asserting that the
  shorthand renders identically to `="true"`, that the omitted/`="false"` cases stay unaffected, and that no console
  errors are logged. No React example is needed, since JSX `hideLabel` already resolves to `true`.
- Updated `stats.json` chunk-size fixture.
- Changelog entry added.


henrijoss:
Follow-up for the clean long-term solution created as #4708: allow `BreakpointCustomizable` objects to be set via property only and drop attribute object support in v5.

That would remove the `string` member from `BreakpointCustomizable<T>`, so Stencil coerces booleans natively again and the workaround in this PR (`parseJSONBoolean()`, `buildResponsiveBooleanStyles()` and the meta-test forbidding `BreakpointCustomizable<boolean | …>` unions) becomes obsolete. Referenced via `TODO: [v5] … see #4708` comments in `breakpoint-customizable.ts`, `jss.ts` and `breakpoint-customizable.spec.ts`.

Stencil's `@AttrDeserialize()` / `@PropSerialize()` (https://stenciljs.com/docs/serialization) was evaluated as an alternative and rejected for now — the decorators only run on attribute → property changes, while all three wrappers set properties, and the DSR/SSR components and jsdom-polyfill are generated separately and would not execute them. Noted as out of scope in #4708.

stephanschroeter (approved):

PR #4702, opened 2026-09-02
Enable typechecking inside our monorepository

### LINKED ISSUE #4186: Add typecheck script
Scope
---
npm typecheck script in every package

`"typecheck": "tsc --noEmit"`

Maybe executable in root package.json with `npm run typecheck --workspaces` all at once

Out of Scope
---

Notes  
---
Share any further information that is relevant for processing.

Acceptance criteria
---
- [ ] Criteria

Subtasks
---
- [ ] Task

kauricre:
### Pull Request Checklist

### Related Issue

Closes #4186


packages/components-react/projects/react-router/tsconfig.json
```diff
@@ -13,7 +13,6 @@
     "moduleResolution": "bundler",
     "jsx": "react-jsx",
     "rootDirs": [".", "./.react-router/types"],
-    "baseUrl": ".",
     "paths": {
       "~/*": ["./app/*"]
     },
```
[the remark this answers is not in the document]
kauricre: Not applicable on TypeScript 6. paths has resolved relative to the tsconfig without baseUrl since TS 4.1, and baseUrl itself is deprecated in TS 6 (re-adding it fails with TS5101).

Source: https://www.typescriptlang.org/tsconfig/#baseUrl

docs/typecheck.md
henrijoss, line 1: We currently have no mechanism in place which keeps these docs in sync, especially when they are full with implementation detail which will become stale very quickly. I would suggest to only explain the mechanism and the conventions of how we do the typechecking here and not all of the concrete details.

tsconfig.paths.json
```diff
@@ -0,0 +1,16 @@
+{
+  "//": "Typecheck-only path mappings, pointing internal specifiers at sibling SOURCE so the repo can be typechecked before anything is compiled. Kept out of shared/src/tsconfig.json because packages/shared/rollup.config.js:88 copies that file into dist, so mappings placed there would ship to consumers pointing at paths the published package does not contain. Extended only from each package's tsconfig.typecheck.json, never its tsconfig.json, so the build keeps resolving from published declarations and a declaration-bundling fault is still caught.",
```
henrijoss: shared dist is not published to consumers

henrijoss:
**Suggestion: give the base tsconfig and the path mappings a proper home in `shared`**

The switch to `shared/src/tsconfig.json` is the right fix — `@porsche-design-system/shared` resolves to `packages/shared/dist` (it's registered as its own workspace entry in the root `package.json`), so a config extending it can't be read in a fresh clone, which is exactly what the new pre-build typecheck pass needs. Keeping a workspace specifier rather than going relative is also right.

I'd like to push it one step further, because two things in the current shape will bite us later.

**1. `src/` is the wrong home for a repo-wide compiler base.** It only lives there so `rollup.config.js:100` can copy it into `dist` as an asset, exposed via the `'./tsconfig.json'` export at `:88`. But after this PR nothing extends the dist copy any more:

```
$ git grep -n "shared/tsconfig.json" -- . | grep -v packages/shared/dist
docs/typecheck.md:97:  ... as `packages/shared/tsconfig.json` does     # prose, different file
```

So both lines are dead and can go — and once they do, there's no reason left for the file to sit in `src/`. Every package currently depends on `shared`'s internal directory layout for no remaining benefit.

**2. The relative references to `tsconfig.paths.json`.** Twenty files, at two different depths:

```
17 × "../../../../tsconfig.paths.json"
 3 × "../../tsconfig.paths.json"
```

Each encodes its own nesting level, so any package move silently invalidates a subset with no single place to fix it — the same problem we're avoiding by not using relative `extends` for the base config.

**Proposed shape**, which fixes both without a new package:

```
packages/shared/
├── tsconfig.base.json     ← moved out of src/
└── tsconfig.paths.json    ← moved from repo root
```

```jsonc
// 31 × package tsconfig.json
{ "extends": "shared/tsconfig.base.json" }

// 20 × tsconfig.typecheck.json
{ "extends": ["./tsconfig.json", "shared/tsconfig.paths.json"] }
```

Same resolution mechanism as today (`node_modules/shared → packages/shared`, present after `npm install`, no build), no relative paths, no reach into another package's `src/`, and the mappings end up sitting next to the sources they point at. Note the mappings then become `./src/index.ts`, `../tokens/src/index.ts` etc., since relative paths resolve against the file that *declares* them — trap #4 in `docs/typecheck.md`.

**To be explicit about what this does not change:** `tsconfig.paths.json` stays. It can't be replaced by workspace resolution, because `@porsche-design-system/shared` must keep meaning `dist` during builds so a declaration-bundling fault still fails, while the typecheck pass needs the same specifier to mean source. One specifier, two meanings per pass — that's what `paths` is for, and it's also why it must stay out of the base config that bundlers read.

**One caveat worth handling while we're in here.** All of this resolves only because `packages/shared/package.json` has no `exports` field, so every subpath is open. Its siblings (`component-meta`, `tokens`, …) all have one, so `shared` growing an `exports` map is a normal future change — and the day it happens, all 51 references break at once. Cheap to pre-empt: there's exactly one deep import today

```
$ git grep -hn "from 'shared/" -- '*.ts' '*.tsx' '*.js' '*.mjs' '*.vue'
shared/src/testing/playwright.vrt
```

so an `exports` map listing `.`, the two tsconfigs and that path makes future additions extend it instead of break it. Otherwise a note in `docs/typecheck.md` is the minimum.

**Also, separately: the `tsconfig.paths.json` header comment gives the wrong reason.** It says mappings placed in the shared base "would ship to consumers pointing at paths the published package does not contain". That can't happen — `@porsche-design-system/shared` is internal-only in full (`docs/public-api.md:130`) and only the four wrappers are published. The real reason is the next sentence: the shared base is the common parent of every bundler-read `tsconfig.json`, so `paths` there would leak into builds. Worth swapping in, and `docs/typecheck.md:72` chains off the same wrong reason with "For the same reason".

Concretely:

1. `git mv packages/shared/src/tsconfig.json packages/shared/tsconfig.base.json`
2. `git mv tsconfig.paths.json packages/shared/tsconfig.paths.json` + rebase the mappings on the new location
3. 31 × `shared/src/tsconfig.json` → `shared/tsconfig.base.json`
4. 20 × `"../../../../tsconfig.paths.json"` → `"shared/tsconfig.paths.json"`
5. `packages/shared/tsconfig.json:2` → `"./tsconfig.base.json"`
6. delete `rollup.config.js:88` and `:100`
7. fix the comment and `docs/typecheck.md:72`

Steps 3 and 4 touch only lines this PR already edits, so the added diff is small.

henrijoss (approved):

PR #4705, opened 2026-09-03
Automate release announcements via Slack

### LINKED ISSUE #4670: Automate release announcements via Slack
Scope
---
When a new release goes through notify the PDS consumers in Slack.

Acceptance criteria
---

kauricre:
### Pull Request Checklist

### Related Issue

Closes #4670


.github/actions/notify-slack-release/action.yml
```diff
@@ -0,0 +1,49 @@
...
+  steps:
+    # Node without `npm ci` across ~35 workspaces; the payload builder has no dependencies.
+    - uses: ./.github/actions/install
+      with:
+        node_modules: 'false'
+    - name: 'Fetch release'
+      shell: bash
+      run: gh api "repos/$GITHUB_REPOSITORY/releases/tags/$INPUT_TAG" > "$RUNNER_TEMP/release.json"
+      env:
+        GH_TOKEN: ${{ inputs.github-token }}
```
henrijoss: Should already be available in composite action github context right? GH_TOKEN: ${{ github.token }}

.github/workflows/notify-release-published.yml
henrijoss, line 1: Do we need this? Any release we do will always be done by the CI. 

If we still want to keep we could have it as a manual dispatch:
```
on:
  workflow_dispatch:
    inputs:
      tag:
        description: 'Release tag to announce, e.g. "v4.7.0".'
        required: true
        type: string
```

kauricre: This is to compensate if the CI ever creates a draft release (for which there is option in the code, but I believe it's unlikely that it will happen). This way, once that draft is manually published we still dispatch an action for notification. It also works for manually creating a release from the UI.

henrijoss (approved):

PR #4707, opened 2026-09-03
Publish storefront to S3 + CloudFront instead of GitHub Pages

### LINKED ISSUE #4369: Deploy Storefront to AWS (S3 + CloudFront) instead of GitHub Pages
> [!NOTE]
> **Phase 0 and Phase 1 are complete and the infrastructure is live** at
> `storefront.pdg-porsche-ds.pdcp.porsche.digital`. Two decisions below have since changed:
> everything runs in **one AWS account (prod)** rather than a dev/prod split, and the CDK repo is
> named **`porsche-design-system/storefront-cdk`**. See the progress comment for the reasoning,
> the replacement security model, and four traps worth knowing before Phase 2.

## Problem

The storefront is deployed by committing the built `packages/storefront/dist` into the separate
`porsche-design-system/porsche-design-system-storefront` repo (`.github/actions/gh-pages`) and force-pushing to `main`.
Every PR (`pr-<number>`), every nightly (`nightly`) and every release (`v<version>`) creates a new folder plus a new
commit containing the full static export.

Consequences:

- **We are running out of disk/repo space.** Measured 2026-09-01: the gh-pages repo is **3.13 GB packed**; the working
  tree is **≥5.93 GB across 47,923 files** (GitHub's tree API truncates the response, so the real figure is ~8–9 GB)
  spread over **27 slugs**. A single recent build is **~412 MB**. Git history retains every past build; `--force` push
  does not reclaim anything. Stale `pr-*` folders are never garbage-collected.
- Checkout of the `gh-pages` repo in CI gets slower with every deploy.
- No control over cache headers, redirects, or security headers (CSP/HSTS) — GitHub Pages serves everything with its
  own defaults.
- Deleting old builds requires history rewriting in a second repo.

## Current state (for reference)

- Deployed by `.github/workflows/deploy.yml` (job `github-pages-storefront`) and `.github/workflows/release.yml`
  (job `github-pages-storefront`).
- Both check out `porsche-design-system/porsche-design-system-storefront` into `./gh-pages` via the
  `GH_DEPLOY_KEY_STOREFRONT` SSH deploy key.
- `.github/actions/gh-pages/action.yml` `rm -rf`s the target folder, copies `packages/storefront/dist` into
  `./gh-pages/<target>`, writes `version.md`, commits and `git push origin main --force`.
- The target slug is resolved once in `contribution.yml` (`context` job): `pr-<number>` / `nightly` / branch name, and
  passed to both build (as `NEXT_PUBLIC_BASE_PATH`) and deploy. Release deploys use `v<version>`.
- `basePath` logic lives in `packages/storefront/src/utils/getBasePath.ts` (`main` → `nightly`), consumed by
  `next.config.ts` and `layout.tsx` (`<base href>`).
- Algolia re-indexing (`npm run updateAlgolia`) piggybacks on the same action, gated by `storefront-algolia`.
- `designsystem.porsche.com` is today a plain `CNAME` to `porsche-design-system.github.io`; GitHub routes by `Host`,
  so the path space is entirely under our control and unchanged by a migration.
- The example apps are **not** on this domain — they live in the separate `porsche-design-system/examples` repo and are
  served at `porsche-design-system.github.io/examples` (GitHub's own domain, no custom domain, no DNS we control).
- Assets already use an S3-compatible object store (Akamai/Linode via rclone, `.github/actions/rclone`).

## Goal

Replace the git-based GitHub Pages deployment with **AWS S3 (origin) + CloudFront (CDN)**, keeping the existing
slug/`basePath` contract (`pr-<n>`, `nightly`, `v<version>`) intact so URLs and the resolved-once `storefront-slug`
logic in `contribution.yml` stay unchanged.

## Environment & constraints (verified 2026-09-01)

Two AWS accounts exist, provisioned by Porsche Digital (PDCP) on an **AWS Control Tower** landing zone. Human access is
**SSO only** (`AWSAdministratorAccess`, `PDCP-AccountDeveloper`, `PDCP-AccountReadOnly`) — no IAM users, no long-lived
keys.

| | `PDG-porsche-ds-dev` | `PDG-porsche-ds-prod` |
| --- | --- | --- |
| Account ID | see private infra repo | see private infra repo |
| CDK bootstrap, `eu-central-1` | ✅ | ✅ |
| CDK bootstrap, `us-east-1` | ❌ **missing** | ❌ **missing** |
| Existing CDK app | `McpCdkStack` (from `porsche-design-system/mcp-cdk`) | same |
| Route 53 public zone | delegated preview zone (id in private infra repo) | delegated prod zone (id in private infra repo) |
| GitHub OIDC provider | exists, PDCP-managed | exists, PDCP-managed |
| Existing budget | `pds-mcp-monthly-dev-cdk` ($100) | `pds-mcp-monthly-prod-cdk` ($100) |
| CloudFront distributions | none | none |
| ACM certificates | none | none |

> **Note**: this repository is public. Account IDs, hosted zone IDs and role ARNs are kept in the private
> infrastructure repository and injected into workflows as GitHub Actions variables — never inlined here.

Constraints that follow:

- **The GitHub OIDC provider must be referenced, not created.** It is created by the Control Tower StackSet
  `StackSet-CustomControlTower-pdcp-iam-github-idp-*`, and an account can only have one provider per URL, so its
  ARN must be passed in as a parameter rather than created. The concrete ARNs live in the private infra repo.
- **There is a working precedent**: the `github-mcp-deployer` role in both accounts trusts
  `repo:porsche-design-system/porsche-design-system:*`. That wildcard is broader than we want (see Decisions).
- **AWS Shield Advanced is subscribed org-wide** (auto-renewing). Protection is *not* automatic —
  resources must be explicitly associated.
- SCPs and tag policies could not be read with the current permissions — **to be confirmed with PDCP**.
- The `McpCdkStack` source lives in the private repo `porsche-design-system/mcp-cdk` (standard CDK layout:
  `bin/`, `lib/`, `test/`, `cdk.json`, Biome + Prettier). It has **no `.github/workflows`** — it is deployed from a
  laptop — so there is **no existing CI deployment pattern to inherit**; this migration establishes it.
- **`us-east-1` is not bootstrapped** in either account. The ACM certificate stack cannot deploy until it is
  (`npx cdk bootstrap aws://<account>/us-east-1`), which must be done locally with SSO admin.

## Decisions

- **AWS CDK (TypeScript), not Terraform.** Both accounts are already CDK-bootstrapped and running `McpCdkStack`, so CDK
  is the house tool; adding Terraform would mean a second IaC tool, a state bucket to own and a BUSL licence question in
  a repo that runs OSS Review Toolkit. CDK is Apache-2.0, needs no state backend, and its `aws-cdk-lib/assertions`
  module lets us assert the IAM invariants below as **automated tests**.
- **The CDK app lives in a new private repo, `porsche-design-system/cdk`** — not in this monorepo. This repository
  is public, and infrastructure code is inherently full of identifiers (account IDs, bucket and role names, exact
  IAM permissions, distribution and zone IDs) that are individually harmless and collectively free reconnaissance.
  A private repo also aligns merge rights with AWS access, and keeps `aws-cdk-lib` (130 MB, 7,350 files) out of
  every contributor's `npm install`. The existing `mcp-cdk` repo is to be folded into `cdk` as a follow-up and then
  archived — one CDK app, many stacks, one bootstrap/OIDC/CI story, one `aws-cdk-lib` bump per dependency run.
  A shared `StaticSite` construct also becomes reusable if the example apps ever leave GitHub Pages.
- **Two distinct IAM roles per account, never conflated.** The *CDK deploy role* (used by the `cdk` repo) needs only
  `sts:AssumeRole` on `cdk-hnb659fds-{deploy,file-publishing,lookup}-role-*` — CloudFormation does the real work via
  `cdk-hnb659fds-cfn-exec-role`, so it does not need `AdministratorAccess`. The *storefront content role* (used by
  this monorepo) gets only `s3:PutObject`/`ListBucket`, a prefix-scoped `s3:DeleteObject` and
  `cloudfront:CreateInvalidation`, with no path to IAM at all.
- **Identifiers are never inlined in this public repo.** Role ARNs reach the workflows as GitHub Actions *secrets*
  (masked in logs, unavailable to fork PRs), not as `vars` and not hard-coded.
- **AWS over Vercel** (the original title's alternative). The asset pipeline is already S3-compatible, the accounts and
  landing zone already exist, the storefront is a pure static export needing no SSR platform, and it avoids onboarding a
  new vendor. Revisit only if SSR/ISR becomes a requirement.
- **Split across the two existing accounts.** `pr-*` previews live in **dev**, `nightly` + `v<version>` in **prod**.
  This turns the "deploy role cannot delete published content" requirement from an IAM condition (one careless edit from
  being wrong) into an **account boundary**, and it means the PR-triggered workflow never holds credentials in the
  account that serves the production domain.
- **Prod CI credentials are append-only for released versions.** The prod role gets `s3:PutObject` on the bucket but
  `s3:DeleteObject` **only** under `nightly/*`; release deploys use `aws s3 sync` *without* `--delete`. A compromised or
  buggy pipeline therefore cannot remove a published version.
- **Use the delegated Route 53 zones.** (Reverses the earlier "no Route 53 hosted zone" assumption — both accounts have
  one.) ACM certificates for `*.pdcp.*` hostnames validate automatically via DNS, so the **entire preview environment
  has no external dependency**. Only `designsystem.porsche.com` still needs corporate DNS, and only at final cutover.
- **PR previews are deleted when the PR is closed, not after a fixed time.** S3 lifecycle rules are time-based only and
  cannot react to a PR closing, so cleanup is event-driven: a workflow on `pull_request: [closed]` (fires on both merge
  and close) removes `s3://<bucket>/pr-<number>/`. A 90-day lifecycle rule is kept purely as a **safety net** for
  orphaned prefixes — without it a single failed run would leak a prefix forever, the exact failure mode we are
  migrating away from.
- **Tighten the OIDC trust condition.** Rather than copying `github-mcp-deployer`'s `repo:<org>/<repo>:*`, scope the
  prod role to a protected GitHub Environment (`repo:porsche-design-system/porsche-design-system:environment:storefront-prod`).
  Fork PRs cannot obtain `id-token: write` today, but the wildcard would also match a `pull_request_target` workflow on
  any branch.
- **Associate both distributions with Shield Advanced.** It is already paid for org-wide; not associating it means
  paying for protection we do not get. It also brings **DDoS cost protection**, which addresses the residual risk below.
- **No AWS WAF managed rule groups for now.** The origin is a static S3 bucket with no compute, forms, auth or user
  input — SQLi/XSS/LFI rule groups protect against classes of attack that cannot occur here, and all content is public
  by design. Note that **WAF is included at no additional charge for Shield-Advanced-protected resources**, so the
  earlier cost rationale no longer applies; the remaining argument is complexity. If billing anomalies appear, add a
  single rate-based rule.
- **Edge coverage: `PriceClass_All`.** The storefront is developed in the EU but consumed worldwide. `PriceClass_100`
  (NA + EU only) would route APAC/LATAM users across an ocean — a regression versus GitHub Pages, which serves globally
  today. Higher per-GB rates apply only to bytes actually delivered from expensive edges, so for a docs site the delta
  is negligible. Cost is controlled via long TTLs, the Budgets alarm and Shield Advanced's cost protection.
- **Bucket is not publicly readable and S3 static website hosting is not used.** The storefront stays publicly
  accessible — this only ensures all traffic goes through CloudFront. S3 website hosting was evaluated and rejected: the
  website endpoint is HTTP-only (no ACM certificate, no custom domain over TLS), cannot emit custom response headers
  (CSP/HSTS), and requires Block Public Access to be disabled. Its only advantage — built-in index-document handling —
  is replaced by a small CloudFront Function we need anyway.
- **No URL redirects are needed for `designsystem.porsche.com`.** It is a plain CNAME to `porsche-design-system.github.io`
  and GitHub routes by `Host`, so repointing DNS preserves the path space exactly. What *does* need reproducing is
  GitHub Pages' trailing-slash `301` (`/foo` → `/foo/`), in the CloudFront Function.

### Infrastructure (one CDK construct, instantiated twice)

- S3 bucket as origin, reachable **only** by CloudFront via Origin Access Control (OAC); S3 Block Public Access stays
  enabled. Region `eu-central-1`.
- Key layout mirrors today's folders: `s3://<bucket>/<slug>/…` where `<slug>` is exactly `inputs.storefront-slug`.
- Cache policies: hashed Next.js assets `immutable, max-age=31536000`; HTML `no-cache`. The deploy action sets matching
  `Cache-Control` on the objects themselves via a two-pass `aws s3 sync`.
- Response headers policy for CSP / HSTS / `X-Content-Type-Options`.
- CloudFront Function for directory-index (`/foo/` → `/foo/index.html`) **and** the trailing-slash `301` — needed
  because the S3 REST endpoint, unlike the website endpoint, has no built-in index document handling. Plain JS, so it is
  unit-testable with Vitest.
- `PriceClass_All`; `shield.CfnProtection` on the distribution.
- Lifecycle rule on `pr-*` prefixes with a 90-day expiry as an orphan safety net.
- ACM certificate in **`us-east-1`** — a hard AWS constraint, CloudFront can only attach certificates from N. Virginia.
  Only the certificate lives there; no content or request data. Requires a second CDK stack plus
  `crossRegionReferences: true`.
- Budgets alarm + SNS topic per account, following the existing `McpCdkStack` pattern:
  `pds-storefront-monthly-{dev,prod}-cdk`.
- If CloudFront standard logging is enabled, the log bucket lives in `eu-central-1` with a short retention — access logs
  contain client IPs and are the only place personal data lands.

| | dev account | prod account |
| --- | --- | --- |
| Serves | `pr-<n>` | `nightly`, `v<version>` |
| Hostname | `storefront.<delegated-preview-zone>` | `designsystem.porsche.com` |
| ACM validation | automatic via Route 53 | 1 corporate DNS request |
| CI role may delete | under `pr-*` | under `nightly/*` only |

### Pipeline

- New `.github/actions/s3-deploy` composite action replacing `.github/actions/gh-pages`:
  - two-pass `aws s3 sync ./packages/storefront/dist s3://<bucket>/<slug>/` — `_next/static/*` with
    `immutable, max-age=31536000`, everything else with `no-cache`
  - `--delete` only where the role permits it (`pr-*`, `nightly/*`)
  - write `version.md` with `GITHUB_SHA` (keep parity)
  - CloudFront invalidation for `/<slug>/*` (a wildcard counts as **one** path; 1,000/month are free)
  - keep the conditional Algolia step (`npm run updateAlgolia` with `P_CURRENT_BRANCH=<slug>`)
- New `.github/workflows/cleanup-preview.yml` on `pull_request: [closed]`:
  - `aws s3 rm --recursive s3://<bucket>/pr-${{ github.event.pull_request.number }}/`
  - CloudFront invalidation for `/pr-<n>/*`
  - same fork/bot gating as the deploy job, since it assumes the AWS role
- GitHub OIDC role assumption; no long-lived AWS credentials anywhere.
- Update `deploy.yml` and `release.yml`: drop the second `actions/checkout` of `porsche-design-system-storefront` and
  the `gh-deploy-key-storefront` secret plumbing (also in `contribution.yml`).
- Add a PR comment / deployment status with the preview URL.

### Phase 0 — prerequisites (no external blockers)

- [ ] Ask PDCP: active SCPs, tag policies, any mandated deployment pipeline, and the house pattern for GitHub OIDC roles
- [x] Create the private `porsche-design-system/cdk` repo (conventions copied from `mcp-cdk`; Vitest rather than Jest,
      to match the monorepo)
- [x] Record account IDs, hosted zone IDs and OIDC provider ARNs there — they must not live in this public repo
- [x] `npx cdk bootstrap aws://<account>/us-east-1` for both accounts, locally with SSO admin, matching the
      `--cloudformation-execution-policies` already used for `eu-central-1`
- [ ] Open a follow-up issue to migrate `mcp-cdk` into `cdk` and archive it

### Phase 1 — preview environment in dev (no external dependencies, fully shippable on its own)

- [x] Scaffold the CDK app in `porsche-design-system/cdk` (`bin/`, `lib/`, `test/`, `cdk.json`)
- [x] CDK CI: `cdk diff` on PRs (read-only role, posted as a PR comment), `cdk deploy` to dev on merge to `main`,
      prod behind a protected GitHub Environment with required reviewers. Local `cdk deploy` is allowed against dev
      only — never prod, except break-glass via SSO admin followed by a PR
- [x] Use `HostedZone.fromHostedZoneAttributes` rather than `fromLookup`, to avoid the lookup role and a stale
      `cdk.context.json`
- [x] Storefront construct: S3 + OAC + CloudFront + cache/response-headers policies + CloudFront Function + lifecycle
      rule + Budgets alarm + Shield protection
- [x] ACM certificate + Route 53 record in the delegated preview zone (auto-validated)
- [x] IAM deploy role referencing the existing OIDC provider, scoped to `pr-*`
- [x] Vitest assertion tests: role cannot delete outside `pr-*`; bucket blocks all public access
- [x] Vitest tests for the CloudFront Function (index resolution + trailing-slash 301)
- [ ] `.github/actions/s3-deploy`
- [ ] `.github/workflows/cleanup-preview.yml`
- [ ] Wire `deploy.yml` PR previews to dev; verify end-to-end on a test PR (deploy → close → prefix gone)

### Phase 2 — production

- [ ] Instantiate the construct in prod with the append-only release policy
- [ ] Request the ACM DNS-validation record for `designsystem.porsche.com` from the corporate DNS owners (**permanent** —
      ACM re-validates through it on auto-renewal)
- [ ] Add `designsystem.porsche.com` as an alternate domain name on the distribution *before* DNS points at it —
      CloudFront rejects an unknown `Host`
- [ ] Migrate existing published versions (`v0`–`v4.x`, 21 version slugs) and `nightly` from the gh-pages repo to S3
- [ ] Update `release.yml` to deploy to prod
- [ ] `.github/actions/is-storefront-version-published`: **no logic change needed** — it probes the public URL, not GH
      Pages. Update the stale `description`/step name, and verify unpublished versions still return non-200. OAC returns
      **403**, not 404, for missing keys: do **not** add a SPA-style `403 → /index.html` custom error response, or every
      version would report as published and the release job would silently skip the storefront deploy

### Phase 3 — cutover and decommission

- [ ] Lower the DNS TTL a day ahead, then repoint the `designsystem.porsche.com` CNAME
- [ ] Verify all version paths, `nightly`, trailing-slash redirects and the e2e/sitemap tests against the new deployment
- [ ] Remove `.github/actions/gh-pages` and the `gh-deploy-key-storefront` secret
- [ ] Archive/shrink or delete the `porsche-design-system-storefront` repo (keep intact until verified — it is the
      rollback path)
- [ ] Document in `packages/storefront/AGENTS.md` and `docs/release.md`

## Open questions

- **Do the example apps move at all?** They are in the separate `porsche-design-system/examples` repo on
  `porsche-design-system.github.io/examples` — GitHub's domain, no DNS we control — and are unaffected by the storefront
  repo's size problem. If they do move, `GITHUB_PAGES_BASE` in
  `packages/storefront/src/components/common/WebsiteViewer.tsx` and `packages/storefront/tests/e2e/fixtures/sitemap.json`
  need updating, and old URLs can only be redirected by keeping that repo's Pages site alive with redirect stubs —
  CloudFront cannot help, since those requests never reach it. Recommendation: leave them where they are.
- **Is `getBasePath.ts` still needed?** The slug is already resolved once in `contribution.yml`; the only logic left in
  the util is the `main` → `nightly` mapping. Either keep it as is, or move that mapping into CI and delete it.
- **Should `nightly` live in dev instead of prod?** It is a consumed, linked artifact, which argues for prod — but it is
  also the only prefix requiring `--delete` in prod.

## Cost

Measured against the current ~9 GB of content:

| Item | Estimate |
| --- | --- |
| S3 storage (~9 GB, `eu-central-1`) | ~$0.22/month |
| Growth (~12 releases/yr × ~400 MB) | +~$0.12/month per year |
| CloudFront egress / requests / function invocations | $0 — within the perpetual free tier |
| Invalidations (`/<slug>/*` = 1 path; ~150/month) | $0 — 1,000/month free |
| ACM certificate | $0 |
| Shield Advanced | already paid org-wide |

**Under $1/month**, versus the current cost of cloning a 3.13 GB repo in every deploy job.

## Out of scope

- Akamai/Linode CDN asset sync (`akamai-cdn-assets`) — already object-storage based, unchanged.
- SSR / Lambda@Edge — storefront remains a Next.js static export.
- Changing the `basePath` / slug scheme.
- AWS WAF managed rule groups (see Decisions).

## Acceptance criteria

- Storefront served via CloudFront over TLS; `nightly`, `pr-<n>` and `v<version>` all resolve.
- All previously published version URLs return 200 at their **unchanged** paths after the DNS repoint.
- `/foo` still `301`s to `/foo/`, matching current GitHub Pages behaviour.
- An unpublished version path returns non-200 (today: 404), so the release publication check stays correct.
- Deploys run from CI without long-lived AWS credentials.
- Closing or merging a PR removes its `pr-<n>/` prefix from S3; verified end-to-end on a test PR.
- The preview deploy role cannot delete anything outside `pr-*`, **and has no access to the prod account at all**.
- The prod deploy role cannot delete published version prefixes.
- Both invariants above are covered by `aws-cdk-lib/assertions` tests in the `cdk` repo's CI, so a regression fails a
  pull request rather than being discovered in production.
- Storage no longer grows unboundedly.
- Fork-PR safety preserved: deploy and cleanup jobs stay gated exactly as deploy is today in `contribution.yml`.
- Algolia indexing still runs only for `main` + `v*` push events.
- Storefront e2e/sitemap tests pass against the new deployment.
- A Budgets alarm is active in each account and the distributions are associated with Shield Advanced.

stephanschroeter:
Closes #4369 (partially — the DNS cutover follows separately).

## Why now

The GitHub Pages content repository is **3.3 GB**, well past the 1 GB published-site limit, and its
builds have been **erroring since 2026-09-02**:

```
2026-09-03T10:59:45Z  errored  Page build failed.
2026-09-02T12:57:10Z  errored  Page build failed.
```

So `designsystem.porsche.com` is already frozen on a week-old nightly and 404s `/v4.7.0-beta.0/`.
Pushing to it is a no-op, which is why this replaces GitHub Pages in one cut rather than
dual-publishing. The content repository is left untouched as the rollback path.

## What changes

| Event          | Slug          | Role                          | `--delete` | Algolia |
| -------------- | ------------- | ----------------------------- | ---------- | ------- |
| `pull_request` | `pr-<n>`      | `storefront-preview-deployer` | yes        | no      |
| `push` main    | `nightly`     | `storefront-release-deployer` | yes        | yes     |
| `push` `v[0-9]`| `v4`          | `storefront-release-deployer` | no         | yes     |
| release        | `v<version>`  | `storefront-release-deployer` | no         | yes     |

- **New `.github/actions/s3-deploy`** reproduces the one-off migration exactly — same three passes,
  two cache classes, four content-type overrides. Divergence would republish the 80k existing
  objects with different metadata. The `.vtt` override is functional: browsers refuse a caption
  track that is not `text/vtt`.
- **`.github/actions/gh-pages` deleted**; `deploy.yml` and `release.yml` publish to S3 and
  invalidate CloudFront.
- **New `cleanup-preview.yml`** removes `pr-<n>/` when a pull request closes — what stops the bucket
  repeating the unbounded growth that broke Pages.

## Two things that are not obvious

**Previews deliberately declare no GitHub Environment.** The preview role trusts the bare
`repo:…:pull_request` OIDC subject; declaring an environment would change the subject to
`…:environment:<name>` and the role would refuse the token. Only the release jobs use
`environment: storefront`, whose name _is_ that role's trust condition and whose branch policy
(`main`, `v[0-9]`, `v[0-9][0-9]`) mirrors `on.push.branches`.

**`is-storefront-version-published` had to move hosts.** Its `== 200` logic is unchanged, but it
probed `designsystem.porsche.com` — a host that no longer receives deploys — so it would have
reported every new release as unpublished forever.

Algolia moves across untouched: it indexes relative paths under a slug-named index, so it was never
host-bound and needs nothing at cutover.

## Validation done

- `aws s3 sync --delete` combined with `--exclude` verified against the real bucket: excluded
  destination keys are never deletion candidates, and each pass prunes only what it owns.
- The action's upload script was **executed for real** against a scratch prefix; every file uploaded
  exactly once, and all ten content-type / cache-control combinations match the migrated objects.
- Both IAM trust policies read back from the live account.
- Branch-policy patterns checked against Ruby `File.fnmatch` — `v-experimental` and `vNext` are
  blocked, `v4` and `v10` allowed.
- Migration verification gate re-run: **all checks passed**, and there is no drift — the content
  repository's only new commits are `pr-*` previews, which the migration excludes by design.

## Follow-ups

DNS cutover (ACM validation record from corporate DNS, certificate replacement), then flip
`AWS_STOREFRONT_BASE_URL`, drop `GH_DEPLOY_KEY_STOREFRONT` and archive the content repository.

No changelog entry: nothing here is reachable through a published subpath.


.github/workflows/cleanup-preview.yml
```diff
@@ -0,0 +1,64 @@
...
+        with:
+          # `storefront-preview-deployer` may only touch `pr-*`, so a bug here cannot reach
+          # `nightly` or a published version. Its OIDC trust condition is the bare
+          # `repo:…:pull_request` subject, which is why this job declares no environment.
+          role-to-assume: ${{ secrets.AWS_STOREFRONT_PREVIEW_ROLE_ARN }}
+          aws-region: eu-central-1
+      - name: 'Remove preview'
+        run: |
+          set -euo pipefail
+          aws s3 rm --recursive "s3://${BUCKET}/pr-${PR_NUMBER}/"
```
stephanschroeter, line 76: What happens when the pr was never deployed to S3 but gets removed when the PR gets closed? Same question to cloudefront invalidation.

.github/workflows/cleanup-preview.yml
```diff
@@ -0,0 +1,64 @@
...
+  cleanup:
+    name: 'Cleanup'
+    # Gated exactly as the deploy job in contribution.yml: secrets and the OIDC token are not
+    # available to fork pull requests, and bot-authored pull requests must not reach the AWS role
+    # even when approved.
+    #
+    # We gate on the PR *author* (`github.event.pull_request.user.login`), NOT on `github.actor`:
+    # `github.actor` is whoever triggered the current run, so it does not reliably identify an
+    # automation-authored pull request. The author is stable regardless of who closes it.
+    if: >-
```
stephanschroeter, line 49: In terms of security, what happens when a fork PR removes these lines and creates a PR against our PDS repo? Will the job be executed? If not what exactly prevents it?

.github/workflows/contribution.yml
```diff
@@ -114,7 +118,10 @@ jobs:
     secrets:
       linode-aws-access-key-id: ${{ secrets.LINODE_AWS_ACCESS_KEY_ID }}
       linode-aws-secret-access-key: ${{ secrets.LINODE_AWS_SECRET_ACCESS_KEY }}
-      gh-deploy-key-storefront: ${{ secrets.GH_DEPLOY_KEY_STOREFRONT }}
+      aws-storefront-preview-role-arn: ${{ secrets.AWS_STOREFRONT_PREVIEW_ROLE_ARN }}
```
stephanschroeter, line 138: explain where those secrets are defined in AWS

stephanschroeter:
Thanks — all three answered below, and the two Copilot findings are fixed in 7c7eed4.

---

### `cleanup-preview.yml:33` — could a fork remove those lines and run the job?

It can remove them, and that changes nothing. This was the most important question here, because the `if:` gate is **not** what stops a fork — I had implied otherwise, so I've rewritten the comment to say what actually holds.

A fork pull request runs the workflow file from **its own head branch**, so an attacker can delete the condition. Two platform guarantees hold regardless:

1. > "With the exception of `GITHUB_TOKEN`, secrets are not passed to the runner when a workflow is triggered from a forked repository."

   So `secrets.AWS_STOREFRONT_PREVIEW_ROLE_ARN` is empty and there is no role to assume.

2. > "You can use the `permissions` key to add and remove read permissions for forked repositories, but typically you can't grant write access."

   So `id-token: write` is downgraded to read and **no OIDC token can be minted at all**. (The documented exception is the "Send write tokens to workflows from pull requests" setting, which applies to forks of *private* repositories — not this one. Repo default is already `default_workflow_permissions: read`.)

This is worth spelling out because **the OIDC subject would otherwise match**: for a fork pull request the `repository` claim is still the *base* repository, so the subject really is `repo:porsche-design-system/porsche-design-system:pull_request`. The role's trust policy alone would not turn a fork away — the absence of a token is what does.

So the condition is defense in depth plus noise reduction: it stops the job from starting and failing confusingly, and it keeps **bot-authored** pull requests away from the role — those *do* run in this repository and would otherwise have credentials, which is the case the `if:` genuinely earns its place.

### `cleanup-preview.yml:56` — what if the PR was never deployed to S3?

Both commands are no-ops rather than errors. Verified against the live bucket and distribution:

```
$ aws s3 rm --recursive s3://pds-storefront/pr-does-not-exist-99999/
exit=0                                    # no keys, nothing printed

$ aws cloudfront create-invalidation --distribution-id … --paths "/pr-does-not-exist-99999/*"
I99WP270KKOSMGYR97QMC40S43   InProgress   # accepted, completes normally
```

So a pull request whose run was cancelled before the deploy job, or whose tests failed, closes cleanly. The only cost is one invalidation path against the free 1,000/month, which the PR volume here does not come close to. I added a comment rather than a guard — a guard would be more code than the thing it protects.

### `contribution.yml:121` — where are those secrets defined in AWS?

Account **473157801658**, `eu-central-1`, provisioned by the private `porsche-design-system/storefront-cdk` repository, which is the source of truth. Now documented inline; summary:

| Secret | AWS resource | Trusted OIDC subject | Write | Delete |
| --- | --- | --- | --- | --- |
| `AWS_STOREFRONT_PREVIEW_ROLE_ARN` | `storefront-preview-deployer` | `repo:…:pull_request` | `pr-*` | `pr-*` |
| `AWS_STOREFRONT_RELEASE_ROLE_ARN` | `storefront-release-deployer` | `repo:…:environment:storefront` | `nightly/*`, `v*` | **`nightly/*` only** |
| `AWS_STOREFRONT_BUCKET` | origin bucket | — | — | — |
| `AWS_STOREFRONT_DISTRIBUTION_ID` | CloudFront distribution | — | — | — |
| `AWS_STOREFRONT_BASE_URL` | public hostname | — | — | — |

Everything is in one AWS account, so there is no account edge protecting published content — those prefix conditions **are** the boundary, and `storefront-cdk`'s `test/iam.test.ts` pins them so a careless edit fails a pull request. Both roles are assumed via OIDC, so unlike the Linode/Akamai pair above there are no long-lived AWS keys for the storefront.

Both trust policies read back from the live account rather than from the CDK source:

```
storefront-preview-deployer  sub = repo:porsche-design-system/porsche-design-system:pull_request
storefront-release-deployer  sub = repo:porsche-design-system/porsche-design-system:environment:storefront
```

That is also why the preview job declares **no** environment: adding one would change the subject to `…:environment:<name>` and the role would refuse the token.

---

### Copilot findings

- **No `curl` timeout** — real, and the worst failure mode in the diff: a stalled connection would hang the release pipeline indefinitely. Now `--connect-timeout 10 --max-time 30 --retry 2`, bounded at ~40 s against a black-holed host instead of forever. It fails *closed*: an unreachable host yields `000`, which reads as "not published" and deploys again — the sync is idempotent and never deletes, so that is the safe direction.
- **Unquoted `SOURCE`** — quoted for consistency. (Strictly it was already safe: `[[ ]]` does no word splitting or globbing, unlike `[ ]`.)

### Verification so far

The first run on this PR is fully green, including the parts that could only be tested in CI:

- `Deploy / Storefront (Preview)` **success** — the OIDC exchange works.
- `Deploy / Storefront (Release)` correctly **skipped** on a `pull_request` event.
- 3,943 objects published; `/pr-4707/` and nested routes return 200.
- Live metadata spot-checked: `.vtt` → `text/vtt` (the functional one), `_next/static` → `public, max-age=31536000, immutable`, documents → `no-cache`, `version.md` → `text/markdown` and stamped with the commit SHA.

The one path still unexercised is cleanup-on-close — which merging this PR will itself trigger, since `pull_request: closed` fires on merge too.

.github/workflows/deploy.yml
```diff
@@ -144,34 +153,79 @@ jobs:
...
-          # Algolia indexing flag is also resolved once in the caller
-          # (contribution.yml → context job) and passed through.
+          slug: ${{ inputs.storefront-slug }}
+          bucket: ${{ secrets.aws-storefront-bucket }}
+          distribution-id: ${{ secrets.aws-storefront-distribution-id }}
+          # Only `nightly` is prunable, and only because the role permits it. A push to a `v*` branch
+          # republishes the major-version alias without removing anything: published content is
+          # append-only so that a buggy pipeline cannot unpublish a release. Orphans left behind are
+          # unreachable unless something links to them directly.
+          delete: ${{ inputs.storefront-slug == 'nightly' }}
```
stephanschroeter: Why not deleting e.g. a v3, v4 deploy?

stephanschroeter: I still think it has to be deleted always because when publishing e.g. /v4 then it's a completely new build like nightly. Imagine a new stable v4 version is released on npm, e.g. 4.7.0. Then we have to update the deployed /v4. Therefore, I think it's necessary to delete the previous /v4 build before publishing a new one. Just like we do with nightly.

stephanschroeter:
You're right, and I've changed it — `/v4` is now replaced rather than merged into.

I had the line in the wrong place: I drew it between `nightly` and everything else, when it actually belongs between **moving** and **pinned** slugs.

| | Examples | Behaviour | `--delete` |
| --- | --- | --- | --- |
| Moving | `nightly`, `v4`, `v10` | rebuilt wholesale every deploy | **yes** |
| Pinned | `v4.7.0`, `v4.7.0-beta.0` | published once, immutable | no |

`nightly` and `v4` are the same kind of thing — an alias that a new build replaces — and GitHub Pages treated them identically, removing the target folder before copying. Without `--delete`, a page removed from the storefront would stay reachable under `/v4/` forever.

The invariant worth protecting is narrower than I'd implemented it: a buggy or compromised pipeline must not be able to unpublish a **released version**. That still holds, by policy and by command — `release.yml` publishes pinned versions without `--delete`, and the role cannot delete under `v4.7.0/` at all.

## ⚠️ Merge order

`delete: 'true'` alone would fail with `AccessDenied`: the release role's delete prefixes are currently `nightly/*` only. That needs **[storefront-cdk#6](https://github.com/porsche-design-system/storefront-cdk/pull/6)** applied first.

**storefront-cdk#6 → CI applies the IAM change → then this PR.** The other order fails a `v4` deploy.

## Why the IAM patterns look odd

IAM resource wildcards are only `*` and `?`, with no character classes — `v[0-9]` isn't available. So the new prefixes are `v?/*` and `v??/*`, covering `v0`–`v99`.

They miss `v4.7.0/…` for one reason only: `?` cannot match the `/` that must follow, and the character after the digit is a `.`. Subtle enough that I didn't want to rely on reading it, so `test/iam.test.ts` now asserts the resulting **decisions** over real key shapes rather than the spelling of the patterns. Cross-checked against `aws iam simulate-custom-policy` on the live account:

```
nightly/index.html         allowed        v4.7.0/index.html         implicitDeny
v4/index.html              allowed        v4.7.0-beta.0/index.html  implicitDeny
v10/index.html             allowed        v3.29.0/assets/a.png      implicitDeny
v0/core/x.js               allowed        pr-4707/index.html        implicitDeny
```

Reverting `deletePrefixes` to `['nightly/*']` fails four of those, so the test is load-bearing. `cdk diff` touches exactly two resources: the release role and its default policy.

One thing worth noting separately, since it's adjacent but not fixed here: `/v4` is only updated by a push to the `v4` **branch**. A release cut from `main` publishes `/v4.7.0/` and updates `/nightly/`, but does not move `/v4`. That's pre-existing behaviour, unchanged by this PR — happy to look at it separately if it isn't intentional.

PR #4711, opened 2026-09-07
Fix React custom-event target types and expose typed element refs

### LINKED ISSUE #4491: Typing for `onInput` event value PInputNumber
### JS Framework

React

3.34.0

_No response_

_No response_

### Browser version

_No response_

### Reproduction Example

https://stackblitz.com/edit/cou2hl1f?file=src%2FExample.tsx

Intersecting the `HTMLElement` and `PInputNumber` types causes the `value` to be of type `never`. Which seems odd and results in effectively canceling out typescript for checks regarding this value.

Image

### Expected behavior

The type should be correct, in best case scenario without having to manually cast.

### Relevant log output

```shell

```

henrijoss:
## Summary

Closes #4491.

Generate DOM-safe React custom-event targets, named event types, and exported host types consistently for `@porsche-design-system/components-react` and `/ssr`. Migrate the remaining React examples, event demo, current React/Next.js/React Router documentation, generated React snippets, and the storefront search ref.

The bug came from combining React props with a DOM element: `HTMLElement & PInputNumberProps` collapses to `never` because `translate` is a DOM boolean but a React `"yes" | "no"` attribute. JSX props must not define the DOM host contract.

## Public type API

For components with custom events (35 components):

- `PInputNumberProps`: JSX properties, React attributes, and callback props.
- `PInputNumberElement`: DOM host properties plus typed `addEventListener` / `removeEventListener`, including native fallbacks.
- `PInputNumberInputEvent`, `PInputNumberChangeEvent`, `PInputNumberBlurEvent`: complete custom events with a typed host `target`.
- Existing `InputNumberInputEventDetail` and other event-detail types remain unchanged.

The generator separates private property, callback, and event-map interfaces. Only React props merge `BaseProps`, replacing conflicting synthetic callback signatures such as `onBlur`. Both wrapper variants use the same generated host contract for event targets. Wrapper refs retain their existing HTMLElement contract; consumers can opt into the exported host types explicitly.

## Before / after

Named handlers no longer need host casts:

```diff
- import type { InputNumberInputEventDetail, PInputNumberProps } from '@porsche-design-system/components-react';
+ import type { PInputNumberInputEvent } from '@porsche-design-system/components-react';

- const onInput = (e: CustomEvent<InputNumberInputEventDetail>) => {
-   const { name, value } = e.target as HTMLElement & PInputNumberProps;
+ const onInput = (e: PInputNumberInputEvent) => {
+   const { name, value } = e.target;
    setForm((prev) => ({ ...prev, [name]: value }));
  };
```

Inline JSX handlers infer the event without an annotation:

```tsx
const [value, setValue] = useState<PInputNumberProps['value']>('');

<PInputNumber
  name="quantity"
  label="Quantity"
  value={value}
  onInput={(event) => setValue(event.target.value)}
/>;
```

Explicitly typed object refs can opt into the host and custom-event listener types:

```tsx
const inputRef = useRef<PInputNumberElement>(null);
<PInputNumber ref={inputRef} name="quantity" label="Quantity" />;

inputRef.current?.addEventListener('input', (event) => {
  const value = event.target.value;
  // value: string | number | null | undefined
});

<PInputNumber
  name="quantity"
  label="Quantity"
  ref={(element: PInputNumberElement | null) => {
    // Explicit annotation; unannotated callback refs still infer HTMLElement | null.
    inputRef.current = element;
  }}
/>;
```

Imports above come from `react` and `@porsche-design-system/components-react`; use `@porsche-design-system/components-react/ssr` for the same PDS types in SSR integrations. The advanced React documentation includes complete examples and listener cleanup.

## Compatibility and scope

**Ref typings are unchanged; automatic concrete ref typing is deferred to V5 (#4712).** Callback refs infer `HTMLElement | null`, `ComponentRef<typeof PInputNumber>` remains `HTMLElement`, and existing object/forwarded refs keep working. No temporary compatibility helper is introduced. Consumers can opt into accurate DOM properties and typed listeners with `createRef<PInputNumberElement>()`, `useRef<PInputNumberElement>(null)`, or an explicitly annotated callback ref. Accepting existing native-element annotations does not make the PDS host an inner native input. V5 will introduce concrete callback-ref inference and stricter object/forwarded refs together.

Custom `onBlur` callbacks now describe the actual PDS custom event rather than an intersection with React's synthetic focus callback; use the concrete event type instead of a React `FocusEvent`.

Existing `CustomEvent<...EventDetail>` handlers remain assignable, but explicitly choosing that annotation retains its weak target type. Separately declared unannotated functions cannot infer their parameter types backwards from a later JSX assignment.

No form value coercion or payload changes: input values retain their existing unions, `event.target` is the PDS host, and `event.detail.target` is still the original native event target. `currentTarget` is not narrowed. Angular and Vue APIs are unchanged. Components without custom events retain their previous element/ref coverage.

DOM-property optionality remains conservative; V5 follow-up #4712 proposes runtime-validated optionality, complete element/method coverage, shared public contracts, framework-native conventions, and a naming/export migration strategy across all four packages.

The changelog documents the new public types, the target-typing fix, the unchanged ref contract, and the custom-`onBlur` breaking-fix migration. Example/documentation changes and internal generator refactoring are not separate changelog entries.

## Validation

- Wrapper generation/build and React / SSR wrapper typechecks passed.
- Generator/parser tests, strict published-package consumer smoke cases for both entry points, ref synchronization tests, SSR input-number rendering, React demo typecheck, and targeted React snippet-generator tests passed.
- Follow-up 0a835c17c3 fixes the checkbox-story values that failed both CI builds (`disabled: true`, `tabIndex: -1`). The exact skills typecheck gate, storefront typecheck, and 16 React snippet-generator tests now pass locally. CI has been retriggered; its new result is pending. The follow-up also replaces deprecated `MutableRefObject` usages with `RefObject` and addresses the snippet-generator review findings.


henrijoss:
## Agreed release sequence

Because this PR contains type-level breaking changes, **do not merge it before the next stable release is completed**.

1. Complete the next stable release without this PR.
2. Afterwards, merge this PR for a beta release.
3. Ask the teams/consumers to try the beta and confirm whether the typing changes and required migrations are acceptable.
4. Gather and address their feedback before proceeding toward a stable release containing these changes.

The beta feedback step is intentional; acceptance of these changes should not be assumed in advance.

packages/components/scripts/wrapper-generator/ReactWrapperGenerator.ts
```diff
@@ -56,15 +69,54 @@ export class ReactWrapperGenerator extends AbstractWrapperGenerator {
   public generateProps(component: TagName, rawComponentInterface: string): string {
     const genericType = this.inputParser.hasGeneric(component) ? '<T>' : '';
     const propsName = this.generatePropsName(component) + genericType;
+    const { properties, events } = this.parseComponentProps(component, rawComponentInterface, genericType);
+
```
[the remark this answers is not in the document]
henrijoss: This was addressed in ef9993ee53: the Unreleased changelog now includes the new public types, the target typing fix, and breaking-change migration entries for refs and custom onBlur callbacks. No additional entries are needed for the internal helper cleanup or storefront examples.

packages/storefront/src/utils/generator/generateReactMarkup.ts
```diff
@@ -125,8 +125,12 @@ export const generateReactControlledScript = (
...
       if (eventValueKey) {
-        eventType && types.push(eventType);
-        return `  const ${eventName} = (e: CustomEvent<${eventType}>) => {
+        const concreteEventType = tagName.startsWith('p-')
+          ? `${pascalCase(tagName.slice(2))}${eventName.slice(2)}Event`
+          : undefined;
+        const importedType = concreteEventType ?? eventType;
+        importedType && types.push(importedType);
+        return `  const ${eventName} = (e: ${concreteEventType ?? `CustomEvent<${eventType}>`}) => {
     set${pascalCase(prop)}(${negateValue ? '!' : ''}e.detail.${eventValueKey});
```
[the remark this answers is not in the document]
henrijoss: Both edge cases are addressed in the latest push. Event types are deduplicated at the final import aggregation, so duplicates across sibling/nested components are covered too. For a non-PDS detail-based handler without eventType, generation now throws an explicit configuration error rather than emitting CustomEvent<undefined> or hiding the missing contract behind any. Regression cases cover duplicate imports, missing payload types, and explicit payload types. These cases were inherited from the earlier generator rather than introduced by the concrete event names.

packages/components/CHANGELOG.md
```diff
@@ -14,6 +14,27 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0),
...
+  `...EventDetail` types and ref typings are unchanged; use an exported host type explicitly for typed object refs
+  ([#4711](https://github.com/porsche-design-system/porsche-design-system/pull/4711))
+
+### Fixed
+
+- **React**: custom-event handlers in the main and `/ssr` entry points exposed an untyped `target`, requiring unsafe
+  casts such as `HTMLElement & PInputNumberProps` that could resolve to `never`. Inline handlers now infer the host
+  properties without casts; existing property unions and event payloads are unchanged
+  ([#4711](https://github.com/porsche-design-system/porsche-design-system/pull/4711))
+- **Breaking Change** **React**: custom `onBlur` callbacks in the main and `/ssr` entry points conflicted with React's
```
kauricre: This fits better inside a **Changed** section and follows previous breaking change announcements.

packages/components-js/tests/unit/specs/ReactWrapperGenerator.spec.ts
kauricre: If I understand this, we are expecting and manually checking that X number of values should be present in `rawProps` and `props`. Wouldn't it be easier to snapshot those values and therefor we immediately see where the drift is?

packages/storefront/src/utils/generator/generateReactMarkup.ts
```diff
@@ -125,8 +125,15 @@ export const generateReactControlledScript = (
   const eventHandler = eventEntries
     .map(([eventName, { prop, value, eventValueKey, eventType, negateValue, toggleValue }]) => {
       if (eventValueKey) {
-        eventType && types.push(eventType);
-        return `  const ${eventName} = (e: CustomEvent<${eventType}>) => {
+        const concreteEventType = tagName.startsWith('p-')
+          ? `${pascalCase(tagName)}${eventName.slice(2)}Event`
```
kauricre: Is it worth it to have this as a shared helper function? Same pattern appears at packages/components/scripts/wrapper-generator/ReactWrapperGenerator.ts:227 and now we need to maintain it on both places.

packages/components/scripts/wrapper-generator/ReactWrapperGenerator.ts
```diff
@@ -133,6 +181,78 @@ export class ReactWrapperGenerator extends AbstractWrapperGenerator {
...
+        !ts.isTypeReferenceNode(eventType) ||
+        eventType.typeName.getText(source) !== 'CustomEvent' ||
+        eventType.typeArguments?.length !== 1
+      ) {
+        throw new Error(`Expected a CustomEvent callback for ${component}.${key}`);
+      }
+
+      // Match the DOM event name used by useEventCallback. Public event types follow the
+      // component's P-prefixed name; existing event-detail names remain unchanged.
+      const typeName = `${pascalCase(component)}${key.slice(2)}Event${genericType}`;
```
kauricre: ^

PR #4732, opened 2026-09-15
Migrate to one `onSlotchange` writing convention

### LINKED ISSUE #3762: Consistent slotchange listeners
Scope
---
Currently we have different implementations to add a `slotchange` event listener to the `slot` element. We should use one consistent solution which works well. 

Out of Scope
---

Notes  
---
- onSlotchange seems to be called initially as well
- evaluate if we can use `onSlotchange` in jsx

Acceptance criteria
---
- [ ] Criteria

Subtasks
---
- [ ] Task

kauricre:
### Pull Request Checklist

### Related Issue

Closes #3762


packages/components/src/components/multi-select/multi-select/multi-select.tsx
```diff
@@ -429,7 +427,7 @@ export class MultiSelect {
           >
             {!this.hasFilterResults && <NoResultsOption />}
             <slot name="options-status" />
-            <slot />
+            <slot onSlotchange={this.onSlotchange} />
```
stephanschroeter, line 430: Is this comment in several places no issue anymore? `// it would be better to use `<slot onslotchange={() => {}} />` in jsx but that doesn't work reliable or triggers initially when component is rendered via js framework`

kauricre: Yes. The "triggers initially" part is real behavior and still happens. The migration handles it instead of avoiding it and side effects like double scrolling (which I assume was a cause for the left comment which states it "doesn't work reliable") are verified with unit + e2e tests pinning that the initial fire is a no-op. The "doesn't work reliable" part has no linked issue I could find, and the new e2e coverage (dynamic adds, empty mount, DOM reattachment) is green in all three browsers.

Issue #331, opened 2019-10-29
Fix storefront code example for events in angular&react


cstickel:
**Scope**
Adjust RegEx for code examples by event handling

**Notes**
- Example:
https://raw.githubusercontent.com/porscheui/porsche-ui-kit/20ce61910e53c4f6235762ca327b0ec9f2a6f5f0/packages/ui-kit-js/src/components/action/button-icon/button-icon.code.md?token=AAIEAQW4C332QWELRGIF7FC5YE5YQ

**Tasks**
- [ ] Adjust RegEx

// QA
- [ ] 3rd party code review

Issue #324, opened 2019-10-28
Component / Buttons: Disabled and Loading State


marcelbertram:
**Description**  
The disabled and loading state of the Button Regular / Button Small / Button Icon is not coded according to the design specification
 
**Expected behavior**  
Code according to design specification
https://share.goabstract.com/2c2b81bc-7cd2-457d-beed-1837114ce942

Change Code Examples Label of Buttons into "Some label"
![image](https://user-images.githubusercontent.com/42934679/67680976-af1a8800-f98c-11e9-8b5b-b31dd9b18511.png)

**Screenshots**  
![image](https://user-images.githubusercontent.com/42934679/67680608-d91f7a80-f98b-11e9-83c7-caa001c0df38.png)

**Tasks**
- [x] Update Disabled and Loading State for Button Regular/Small
- [x] Update Disabled and Loading State for Button Icon
- [x] Update Docu Example of Button Regular/Icon
- [x] Update vrt fixtures
- [x] Rewrite Component Labeling of Buttons into "Some label"

Issue #237, opened 2019-08-22
Storefront / Add broken link testing for storefront


o-hook:
**Scope**  
Dead links are not cool! An automatic link checker tool should be integrated into storefront which should be integrated into testing environment.

**Notes**  
After some research, this might be a good solution: 
https://matthewsetter.com/writing-tools/npm-broken-link-checker/

**Subtasks**  

// DEV
- [x] Research link checker tools
- -> npm broken link checker does not work with shadow root anchor tags
- [ ] Implement via puppeteer e2e

PR #4736, opened 2026-09-16
Fix polynomial ReDoS findings reported by GitHub Code Scanning

stephanschroeter:
## Motivation

[GitHub Code Scanning](https://github.com/porsche-design-system/porsche-design-system/security/code-scanning) reports four open `js/polynomial-redos` alerts (high severity, CWE-1333):

| Alert | Location |
| --- | --- |
| [#324](https://github.com/porsche-design-system/porsche-design-system/security/code-scanning/324) | `packages/shared/src/utils/convertToReact.ts:5` |
| [#330](https://github.com/porsche-design-system/porsche-design-system/security/code-scanning/330) | `packages/shared/src/utils/convertToReact.ts:29` |
| [#331](https://github.com/porsche-design-system/porsche-design-system/security/code-scanning/331) | `packages/shared/src/utils/convertToReact.ts:32` |
| [#356](https://github.com/porsche-design-system/porsche-design-system/security/code-scanning/356) | `packages/component-meta/src/utils/isDeprecatedComponent.ts:12` |

Running the CodeQL CLI locally reproduced all four and surfaced **six more of the same kind** in `convertToVue.ts` and `convertToAngular.ts` that have not been reported yet (newer query version). Those are fixed here too, so they never show up as new alerts.

The fifth open alert, [#237](https://github.com/porsche-design-system/porsche-design-system/security/code-scanning/237) (`js/incomplete-sanitization`), is **already fixed on `main`** — `packages/components/src/utils/validation/partials/helper.ts` was deleted in `9792dd48cd`. Its instance on `refs/heads/main` is `fixed`; the alert only stays open because of stale long-lived PR branches. Nothing to do.

## Cause

All ten findings share one root cause: an unbounded quantifier can also consume the delimiter that follows it, so a failing match must backtrack over every candidate split. With many restart positions this becomes quadratic.

| Pattern | Problem |
| --- | --- |
| `\s(\S+)="…"` | `\S` also matches `=` and `"` |
| `="{(.*?)}"` / `="'(.*?)'"` | `.` also matches the closing quote |
| `(<input(?:.[^/]*?))>` | `.`/`[^/]` also match `<` |
| `(<([A-Za-z-]+)[^>]*?)>` | `[^>]` also matches `<` and the tag-name chars |
| `@deprecated ([^*\n]*)[\s\S]*?@Component` | the two quantifiers overlap |

## Changes

Each quantifier is narrowed to the characters it is actually meant to cover, and `isDeprecatedComponent()` is split into an `indexOf('@Component')` plus a single non-backtracking regex.

### Drive-by bug fix

Narrowing `[^>]*?` to `[^<>]*` in `transformToSelfClosingTags()` also fixes a real bug. For slotted content that looks like an attribute the old regex matched *across* tag boundaries and swallowed the closing tag:

```diff
  <PSelect label="Select hyphens" name="hyphens">
    <PSelectOption valueAutoStyle="hyphens: auto;"</PSelectOption>
-   <PSelectOption valueNoneStyle="hyphens: none;"</PSelectOption />
+   <PSelectOption valueNoneStyle="hyphens: none;"</PSelectOption>
+ </PSelect>
```

This affects the React code examples of `link-tile-example-hyphens.html` and `button-tile-example-hyphens.html` in the storefront. A regression test was added.

## Verification

- **CodeQL CLI 2.27.0** (`codeql/javascript-queries:Performance/PolynomialReDoS.ql`) against a database of `packages/shared` + `packages/component-meta`: **10 results before → 0 results after**
- **Behaviour** — old vs. new conversion diffed over all **768** HTML files in the repo: the only 6 differences are the closing-tag bug above
- **Generated output** — all React, Next.js, React Router and Angular VRT pages regenerated: **byte-identical**
- **`isDeprecatedComponent()`** — old vs. new over all 85 component `.tsx` files: **0 differences**
- **Timing** on the exact pump strings from the alerts:

  | Helper | before | after |
  | --- | --- | --- |
  | `transformObjectValues` | 1564 ms | 0 ms |
  | `transformInputs` | 1892 ms | 0 ms |
  | `transformToSelfClosingTags` | 4130 ms | 0 ms |
  | `isDeprecatedComponent` | 249 ms | 0 ms |

- Unit tests: `packages/shared` 99 passed, `packages/component-meta` 8 passed
- `biome check` clean

## Changelog

No entry. `@porsche-design-system/shared` and `@porsche-design-system/component-meta` are `"private": true` and are not reachable through any published subpath (see `docs/public-api.md`), so this is not consumer-facing.


stephanschroeter:
## Response to the Copilot review

Three of the five findings were acted on, two are rejected. Details below.

### ✅ Accepted — `indexOf('@Component')` can truncate the message

Correct and reproducible. For a message that mentions the decorator, the slice ended inside the JSDoc:

```
/**
 * @deprecated since v3.0.0, use @Component with shadow: false instead.
 */
@Component({ … })
```

| | result |
| --- | --- |
| before this PR | `[true, "since v3.0.0, use @Component with shadow: false instead."]` |
| first commit (buggy) | `[true, "since v3.0.0, use "]` ❌ |
| now | `[true, "since v3.0.0, use @Component with shadow: false instead."]` ✅ |

Fixed in 307982894a by anchoring on `@Component(`, which only ever matches the Stencil decorator — all 75 component files use it, none has a bare `@Component`. A regression test was added.

I deliberately avoided a line-anchored regex such as `/^\s*@Component\b/m` here: `^\s*` under `/m` reintroduces exactly the many-start-positions-times-unbounded-scan shape this PR is removing.

### ✅ Accepted — missing active Vue coverage

Right, `transformVueInputs()` and `transformVueToSelfClosingTags()` were only referenced from the skipped `convertToVue()` integration suite. Both now have active describe blocks mirroring the React suite, including the tag-boundary regression case.

### ❌ Rejected — `[^"\n]*` "does not actually remove the ReDoS"

This is not correct.

1. **CodeQL disagrees.** The CodeQL CLI (2.27.0, `Performance/PolynomialReDoS.ql`) goes from 10 results to 0 on these files, and the CodeQL run on this PR reports `results_count: 0`.
2. **Measured, it is linear.** `[^"\n]` cannot cross the `"` that every start position requires, so each attempt is bounded by the inter-quote distance:

   | input | `unbindAngularNativeAttributes` | `unbindVueNativeAttributes` |
   | --- | --- | --- |
   | 240 000 chars, 20 000 unterminated attrs | 1 ms | 0 ms |
   | 480 000 chars, 40 000 unterminated attrs | 1 ms | 0 ms |
   | 960 000 chars, 80 000 unterminated attrs | 3 ms | 0 ms |

   Doubling the input doubles the time — that is linear, not polynomial.
3. **The suggested fix would be a regression.** Excluding the single quote changes behaviour for values that legitimately contain one, because the original `(.*?)` matched up to the first `'"`, not the first `'`:

   ```
   input                <p-tag [title]="'it's here'"></p-tag>
   this PR              <p-tag title="it's here"></p-tag>          ✅ same as before
   suggested [^"'\n]    <p-tag [title]="'it's here'"></p-tag>      ❌ no longer unbound
   ```

The same reasoning applies to the identical Vue comment.

henrijoss (approved):

PR #4737, opened 2026-09-16
Harden code scanning: main-only triggers, weekly schedule, and a documented CodeQL category

stephanschroeter:
Follow-up to #4736. While clearing the four alerts that survived their own fix, several problems with `code-scanning.yml` surfaced.

## 1. Scope: `main` only, but now also scheduled

The triggers stay restricted to `main` — the `v*` branches are not maintained, so scanning them would only produce findings nobody acts on. `contribution.yml` still builds and tests them; that is a separate concern.

What is new is a **weekly schedule** (Mondays 03:00 UTC). CodeQL ships new and improved queries independently of this repository, so unchanged code can start failing a query that did not exist yet. Concretely: while working on #4736 the CodeQL CLI found **6 findings in `convertToVue.ts` / `convertToAngular.ts` that GitHub had never reported** — identical code, newer query pack. Without a cron those only surface on the next push to `main`.

## 2. The analysis category, and why four alerts stayed open

CodeQL derives the analysis *category* from `<workflow file path>:<job id>`. GitHub closes an alert only once **every** category that ever reported it reports it as fixed.

The job was renamed `build` → `code-ql` (7196fcd81f) and back → `build` (d6ab7289dd). That left `.github/workflows/code-scanning.yml:code-ql` orphaned, frozen at commit `8410adfb` (2025-07-18) and unable to ever refresh:

| Alert | `…:build` | `…:code-ql` (orphaned) | Alert state |
| --- | --- | --- | --- |
| 324, 330, 331 | ✅ fixed | ⚠️ open @ `8410adfb` | stuck **open** |
| 237 | ✅ fixed 2025-11-05 | ⚠️ open @ `8410adfb` | stuck **open** |
| 356 | ✅ fixed | *never seen* (postdates the rename) | **closed** ✅ |

Alert 237 had been genuinely fixed for ten months. All four were cleared by deleting the six orphaned analyses by hand. The security tab is now at **0 open** (63 dismissed, 18 fixed).

This is now documented at the top of the workflow, since nothing in the file hinted at it.

### What a category change does *not* do

Worth stating explicitly, because an earlier revision of this PR claimed otherwise: **alert identity survives a category change.** Alerts keep their number, gain instances under the new category, and dismissals are not lost.

- Alerts 237 and 324 each had instances under *both* categories — same alert number.
- **51 of the 63** dismissals have `dismissed_at` before `2025-07-15`, i.e. they predate the `code-ql` category entirely, lived through the whole `build` → `code-ql` → `build` cycle, and are still dismissed today.

Only the open→fixed transition is pinned. That is a subtler failure than "alerts get recreated" — and a nastier one, because it stays invisible until somebody fixes a finding and the alert refuses to close.

## 3. Job renamed `build` → `codeql`

`build` was a leftover from the CodeQL starter workflow and describes nothing this job does. `codeql` matches the existing `name: 'CodeQL'`, the upstream spelling (`github/codeql-action`, the `codeql` CLI) and this repository's job-id convention of keeping a product name as one token — `components-react-nextjs`, not `next-js`.

The category is deliberately **not** pinned to keep the old string. An earlier revision did pin it, which kept tracking perfectly continuous but left a category permanently contradicting the job id — an obvious "cleanup" for the next reader, and a worse trap than the one it prevents.

So the category moves to `…:codeql`, verified on this PR before merging:

```
$ gh api .../code-scanning/analyses?ref=refs/pull/4737/merge --jq '.[0] | "results:\(.results_count) category:\(.category)"'
results:0 category:.github/workflows/code-scanning.yml:codeql
```

**The trade-off, stated plainly.** This orphans `…:build`, whose frozen analysis still reports the 63 findings dismissed as "won't fix". Those alerts keep their terminal dismissed state (see above), so the only lasting effect is that they can never *also* flip to "fixed" later. Everything reported from here on is tracked solely under `…:codeql` and closes normally.

Note that fixing or re-dismissing the findings first would not avoid this: a published analysis is immutable and keeps reporting what it saw. Only deleting the analyses removes them — 981 of them here — and that is what GitHub warns "may result in the loss of historical alert data". Doing the switch now, at zero open alerts, is the cheapest it will ever be.

The PR check is still named `CodeQL` (that comes from `name:`, not the job id), so branch protection is unaffected.

## 4. Smaller changes

- **Concurrency group** cancelling superseded `pull_request` runs only. A cancelled run uploads no analysis, so `push`, `schedule` and `workflow_dispatch` must always finish — otherwise the category goes stale, which is the failure mode above.
- Dropped the commented-out Playwright `container:` block and added `timeout-minutes: 30`, consistent with the jobs in `build.yml`.

## Changelog

No entry. CI configuration only, nothing reachable through a published subpath.


stephanschroeter:
## Correction to the PR description

The description above claims that changing the analysis category would "re-report every currently dismissed alert as a new one, silently dropping its `won't fix` state". **That is wrong**, and 66fbf51a21 corrects the note in the workflow accordingly.

Checking it against this repository's own history:

- Alerts **237** and **324** each had instances under *both* `…yml:build` and `…yml:code-ql`. Same alert number across the category change — so a category change does not mint new alerts.
- **51 of the 63** current dismissals have `dismissed_at` earlier than `2025-07-15`, i.e. they predate the `code-ql` category entirely. They lived through the whole `build` → `code-ql` → `build` cycle and are still dismissed today.

So alert identity and dismissals survive a category change. What a rename actually breaks is narrower: the orphaned category never receives another analysis, so its last one is frozen and keeps reporting whatever it saw, and GitHub closes an alert only once *every* reporting category calls it fixed. Only the open→fixed transition is pinned.

That is a subtler failure than I described — and arguably a nastier one, because it stays invisible until somebody fixes a finding and the alert refuses to close. Exactly how alert 237 sat "open" for ten months after its file was deleted.

The second claim, that an explicit `category:` input is equally unsafe, was also too broad. Setting it to a *different* value changes the category; setting it to the *current* value pins it and is precisely the supported way to rename the job safely. The workflow note now documents that escape hatch, including how to verify it on a PR before merging.

Nothing about the actual configuration in this PR changes — only the comment. The job id stays `build` and the category stays `.github/workflows/code-scanning.yml:build`, confirmed on this PR's own run:

```
$ gh api .../code-scanning/analyses?ref=refs/pull/4737/merge --jq '.[] | "results:\(.results_count) category:\(.category)"'
results:0  category:.github/workflows/code-scanning.yml:build
```

stephanschroeter:
## Re-requesting review — the previous one is stale

The earlier review was submitted on `024d7910`, which is three commits behind the current head `f0e8e506`. Its summary ("scans version branches...") describes a design that has since been reverted, so it should not be taken as an approval of what is now being merged.

What changed after that review:

| Commit | Change |
| --- | --- |
| `66fbf51a` | Corrected a factual claim: a category change does **not** drop dismissals |
| `b7ac63d5` | Renamed the job `build` → `codeql`, pinning the category |
| `f0e8e506` | Reverted to `main`-only scanning and **removed** the pin |

Net effect versus the reviewed commit: version-branch scanning is gone (the `v*` branches are not maintained), and the analysis category deliberately moves from `…yml:build` to `…yml:codeql`.

### Worth reviewing carefully

1. **The category move is the one consequential call.** It orphans 981 `…:build` analyses whose frozen snapshot still reports the 63 findings dismissed as "won't fix". Those alerts keep their dismissed state — verified, 51 of them survived the same thing during the `code-ql` era — so the only lasting effect is that they can never *also* flip to "fixed". The alternative (pinning the category) was tried in `b7ac63d5` and rejected because it leaves a category string permanently contradicting the job id.

2. **The `CodeQL — "1 configuration not found"` warning on this PR is expected.** It is GitHub noticing the category moved and being unable to diff against main's `…:build` baseline. Conclusion is `neutral`, CodeQL is not a required status check, and it clears once main's baseline becomes `…:codeql` after merge.

3. **One claim in this PR was wrong and was corrected.** An earlier revision stated that a category change re-reports dismissed alerts as new ones and drops their "won't fix" state. It does not. Please sanity-check the corrected reasoning in the workflow header rather than trusting it.

stephanschroeter:
## Response to the review

Two of the three findings are fixed in 3411b077b9; the concurrency one is rejected.

### ✅ Runbook reference — correct, and a genuine miss on my side

`docs/runbooks/dependency-updates-agent.md:355-356` really did name the commented Playwright block in `code-scanning.yml` as one of the places to keep in sync when bumping `@playwright/test`, and this PR deleted that block. The reference is removed; `contribution.yml` (4 occurrences) and `docker-compose.yml` (2) are now the only places, which matches `grep -rn "mcr.microsoft.com/playwright"`.

### ✅ `workflow_dispatch` escapes the branch filter — correct

`branches:` does not apply to `workflow_dispatch`, so the workflow could be dispatched against `v4` and publish an analysis for a branch this PR deliberately stopped scanning. Guarded now.

One correction to the suggested remedy, though: gating on `github.ref == 'refs/heads/main'` alone would have broken PR scanning entirely, because `github.ref` is `refs/pull/<n>/merge` on a pull request. The event has to be tested first:

```yaml
if: github.event_name != 'workflow_dispatch' || github.ref == 'refs/heads/main'
```

| event | ref | result |
| --- | --- | --- |
| `push` | `refs/heads/main` | runs |
| `pull_request` | `refs/pull/4737/merge` | runs |
| `schedule` | `refs/heads/main` | runs |
| `workflow_dispatch` | `refs/heads/main` | runs |
| `workflow_dispatch` | `refs/heads/v4` | **skipped** |

### ❌ Concurrency — the mechanic is right, the conclusion is not

The description of GitHub's behaviour is accurate: with `cancel-in-progress: false` a *pending* run is replaced when a newer run enters the same group. But that is the desired behaviour here, and the proposed fix would make things worse.

1. **Nothing is lost when a pending run is dropped.** `push`, `schedule` and `workflow_dispatch` all analyse the same ref (`main`) with the same CodeQL query pack. Whichever run survives publishes an equivalent, *newer* analysis. A scheduled run displaced by a push does not skip the "unchanged code, newer queries" check — the push run performs exactly that check, on a newer commit.
2. **Dropping superseded pending runs is the point.** On a burst of pushes to `main` there is no value in analysing each intermediate commit; the group collapses them to the latest. That is the same reasoning as `contribution.yml`.
3. **`github.run_id` would disable the grouping entirely**, since it is unique per run. Every push would then start a concurrent CodeQL analysis, which is redundant work and, with several runs finishing out of order, means the newest-completing analysis is not necessarily the newest commit.

The invariant that actually matters — never cancel a *running* non-PR analysis, because a cancelled run uploads nothing and leaves the category stale — is exactly what `cancel-in-progress: ${{ github.event_name == 'pull_request' }}` already guarantees. That is stated in the comment above the block.

